PR #1, opened 2020-08-03
User entity: class

bori00:
I created a class for storing the User's data

The User class holds the user's data, with the id and email being mandatory, and all the other fields(selfIntroduction, photo and name) being optional. The user's portfolio is set to private by default. 

In order to make the instantiation easier, I applied the builder pattern for the User.

I created some tests to check if the instantiation works as expected.


src/main/java/com/google/sps/user/User.java
```diff
@@ -0,0 +1,81 @@
...
+
+/** Stores the data related to one user */
+public class User {
+    private final String id, email;
+    private final boolean publicPortfolio;
+    @Nullable 
+    private final String name, selfIntroduction, imgUrl;
+
+    public static class Builder {
+        //Required
```
denniswillie: start comment with space and end with period.

src/main/java/com/google/sps/user/User.java
```diff
@@ -0,0 +1,81 @@
...
+        private final String id, email;
+        //Optional
```
denniswillie: start comment with space and end with period.

src/main/java/com/google/sps/user/User.java
```diff
@@ -0,0 +1,81 @@
...
+        private String name, selfIntroduction, imgUrl;
```
denniswillie: Change to final variable? We wouldn't need to set a field more than once.

src/main/java/com/google/sps/user/User.java
```diff
@@ -0,0 +1,81 @@
...
+        private boolean publicPortfolio = false;
```
denniswillie: Same here, change to final variable.

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,89 @@
+package com.google.sps;
+
+import com.google.sps.user.User;
+import org.junit.Assert;
```
denniswillie: We can import assertEquals so that the code can be shorter :)

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,89 @@
...
+    private static final String ID = "userid";
+    private static final String EMAIL = "user@gmail.com";
+    private static final String NAME = "username";
+    private static final String SELF_INTRODUCTION = "I am the user";
+    private static final String IMG_URL = "/img.com";
+
+    @Test
+    public void getId() {
+        User newUser = new User.Builder(ID, EMAIL).build();
+        Assert.assertEquals(newUser.getId(), ID);
```
denniswillie: Here is `assertEquals(result, expected)`, if there's an error, then the console will display something like:
expected : `result`. but was : `expected`
Change it to assertEquals(expected, result) to make it easier to read if a failure occurred. Modify the others also :) thanks.

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,89 @@
...
+    public void getEmail() {
+        User newUser = new User.Builder(ID, EMAIL).build();
+        Assert.assertEquals(newUser.getEmail(), EMAIL);
+    }
+
+    @Test
+    public void getName() {
+        // User with no preset name.
+        User newUser = new User.Builder(ID, EMAIL).build();
+        Assert.assertEquals(newUser.getName(), null);
```
denniswillie: Maybe for each of these test cases, we can add different test methods so if an error occurred, we can easily find the error occurred.

I use this convention for naming the test method (given by Chris):
`<methodUnderTest>_<condition>_<expectedOutcome>`

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+        // User with preset selfIntroduction, and other fields.
+        newUser = new User.Builder(ID, EMAIL).setPublicPortfolio().setName(NAME).addSelfIntroduction(SELF_INTRODUCTION).build();
+        Assert.assertEquals(newUser.getName(), NAME);
+    }
+
+    @Test
+    public void getSelfIntroduction() {
+        // User with no preset selfIntroduction.
+        User newUser = new User.Builder(ID, EMAIL).build();
```
denniswillie: Maybe also add a test where no parameters is given inside the `User.Builder()` ?

src/main/java/com/google/sps/user/User.java
bori00: changed

src/main/java/com/google/sps/user/User.java
bori00: changed

src/main/java/com/google/sps/user/User.java
bori00: You are right about not changing these fields more than once, but I can't change them to final in the Builder class, because I need to assign values to them outside the constructor. It would result in "cannot assign a value to final variable <variable name>" error. 
See [Effective Java](https://effectivejava.googleplex.com/third-edition), Item 2, page 13. It uses the same pattern, and also omits the final keyword. 
Remark that the fields of the actual User class are final.

src/main/java/com/google/sps/user/User.java
bori00: Same as above

src/test/java/com/google/sps/UserTest.java
bori00: Done

src/test/java/com/google/sps/UserTest.java
bori00: Changed. Thank you for the remark!

src/test/java/com/google/sps/UserTest.java
bori00: You mean no parameters for the constructor of User.Builder?
Given that I have one constructor defined, I can't use a default no-argument constructor for User.Builder

src/test/java/com/google/sps/UserTest.java
bori00: Applied naming convention.

src/test/java/com/google/sps/UserTest.java
denniswillie: okay

denniswillie:
LGTM!

src/main/java/com/google/sps/user/User.java
```diff
@@ -0,0 +1,81 @@
+package com.google.sps.user;
+
+import org.jetbrains.annotations.Nullable;
+
+/** Stores the data related to one user. */
+public class User {
+    private final String id, email;
```
tlheng: Define each variable in separated line. It is more readable

src/main/java/com/google/sps/user/User.java
```diff
@@ -0,0 +1,81 @@
...
+    private final boolean publicPortfolio;
```
tlheng: I thought we agree line indentation is 2 spaces?

src/main/java/com/google/sps/user/User.java
```diff
@@ -0,0 +1,81 @@
...
+    @Nullable 
+    private final String name, selfIntroduction, imgUrl;
+
+    public static class Builder {
```
tlheng: Maybe take a look at AutoValue: https://github.com/google/auto/blob/master/value/userguide/builders.md

It is very nice to use library; removes a lots of boiler plate code.

src/main/java/com/google/sps/user/User.java
```diff
@@ -0,0 +1,81 @@
...
+        // Required.
+        private final String id, email;
+        // Optional.
+        private String name, selfIntroduction, imgUrl;
```
tlheng: Maybe use Optional<>

AutoValue can handle this for you

src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+import com.google.sps.user.User;
+import static org.junit.Assert.assertEquals;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class UserTest {
```
tlheng: User class is a POJO class. No need for unit tests.

src/main/java/com/google/sps/user/User.java
bori00: Done.

src/main/java/com/google/sps/user/User.java
bori00: Yes, sorry, I didn't find this setting in the cloud shell editor. 
I switched to intellij now, and changed the indentation.

src/main/java/com/google/sps/user/User.java
bori00: Added `Optional<>`

src/test/java/com/google/sps/UserTest.java
bori00: I understand that these tests are not needed, but they were and they still are useful for me in case of changes like the ones requested for this CL. 
Should I delete this file, or can I leave it there if it's already written?

tlheng: Let's delete this. No need to submit un-needed code. We will have to maintain it later on.

tlheng (approved):

src/main/java/com/google/sps/user/User.java
bori00: Removed `optional<>`, because it's not supported by Datastore.

src/test/java/com/google/sps/UserTest.java
bori00: I deleted it.

src/main/java/com/google/sps/user/User.java
bori00: I will consider AutoValue for later classes.

PR #3, opened 2020-08-05
User entity: database

bori00:
The recent changes enable saving and retrieving the user's data in a database. 

- I created a **UserRepository** interface, which provides functions for accessing the database. 
- I created a **UserRepositoryFactory** class, which let's the user of the UserRepository interface choose the database in which the user's data will be stored, by providing a **RepositoryType**. Currently, the only supported RepositoryType is DATASTORE, which relies on the **Datastore API.** 
- I implemented the UserRepository interface in the **DatastoreUserRepository** class, which handles the storage of the user's data   based on the Datastore API. 

In addition to these, I created the **DatastoreUserRepositoryTest** class, which tests if the saveUser() and getUser() functionalities work correctly for DatastoreUserRepository. This test relies on a mock database, meaning that the data is in fact stored in-memory only. At the end of the tests, tearDown() wipes out all in-memory data.


bori00:
This PR relies in the code sent to the "User entity: class" PR, which was not merged to the master yet.

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
```diff
@@ -0,0 +1,69 @@
...
+    public static final String IMG_URL_PROPERTY = "imgUrl";
+
+    @Override
+    public void saveUser(User user) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        datastore.put(getUserEntity(user));
+    }
+
+    private Entity getUserEntity(User user) {
+        Entity userEntity = new Entity(ENTITY_NAME, user.getId());
```
denniswillie: Change `ENTITY_NAME` to `KIND_NAME`

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
```diff
@@ -0,0 +1,69 @@
...
+        userEntity.setProperty(SELF_INTRODUCTION_PROPERTY, user.getSelfIntroduction());
+        userEntity.setProperty(IMG_URL_PROPERTY, user.getImgUrl());
+        return userEntity;
+    }
+
+    @Override 
+    @Nullable
+    public User getUser(String id) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        Key userKey = KeyFactory.createKey(ENTITY_NAME, id);
```
denniswillie: change to `KIND_NAME` as well.

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
```diff
@@ -0,0 +1,69 @@
...
+        try {
+            return getUserFromUserEntity(datastore.get(userKey));
+        } catch (EntityNotFoundException e) {
+            return null;
+        }
+    }
+
+    @Nullable
+    private User getUserFromUserEntity(Entity userEntity) {
+        if (userEntity==null) {
```
denniswillie: fix space

src/test/java/com/google/sps/DatastoreUserRepositoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+    private static User toSaveUser; 
+    private static User toGetUser; 
+
+    private static UserRepository myUserRepository; 
+
+    private static LocalServiceTestHelper helper;
+
+    @Before
+    public void setUp() {
+        helper = new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
```
denniswillie: make this a private final field?

src/test/java/com/google/sps/DatastoreUserRepositoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+        helper.setUp();
+    }
+
+    @After
+    public void tearDown() {
+        helper.tearDown();
+    }
+    
+    @Test
+    public void saveUserGetUser_existingUser_returnsUserEqualToSavedUser() {
```
denniswillie: Maybe instead of testing 2 methods at the same time, we can create 2 constructors on the `DatastoreUserRepository`, where one of them accepts external-made datastore (specifically for testing) and the other one doesn't accept any arguments, hence, creating its own datastore. Then on the test method, instead of using `myUserRepository.getUser`, we can retrieve the entity from our datastore and check if the datastore only consists 1 entity and the right entity.

src/test/java/com/google/sps/DatastoreUserRepositoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+        myUserRepository.saveUser(toSaveUser);
+        toGetUser = myUserRepository.getUser(ID);
+        assertEquals(toSaveUser, toGetUser);
+    }
+
+    @Test
+    public void getUser_inexistentUser_returnsNull() {
```
denniswillie: We can also make a new test method to test if the `getUser` method works when there is an existing user. Store the entity first in our own datastore without using `saveUser`.

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
bori00: Changed to ENTITY_KIND, as agreed

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
bori00: Changed

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
bori00: Fixed

src/test/java/com/google/sps/DatastoreUserRepositoryTest.java
bori00: Done that. This way I had to remove the initialization from setUp().

src/test/java/com/google/sps/DatastoreUserRepositoryTest.java
bori00: Done. I think the separate constructors are not needed, as long as I use a mock database anyway.

src/test/java/com/google/sps/DatastoreUserRepositoryTest.java
bori00: Done.

denniswillie:
LGTM!

src/main/java/com/google/sps/user/repository/UserRepository.java
```diff
@@ -0,0 +1,13 @@
+package com.google.sps.user.repository;
+
+import com.google.sps.user.User;
+import org.jetbrains.annotations.Nullable;
+import java.util.HashMap;
+
+/** Handles the storage of User data. */
+public interface UserRepository {
+    public void saveUser(User user);
```
tlheng: Interface methods are public by default, no need `public`. Here and below.

src/main/java/com/google/sps/user/repository/UserRepository.java
```diff
@@ -0,0 +1,13 @@
...
+
+    @Nullable
+    public User getUser(String id);
```
tlheng: Maybe add some docs. At least explain when this returns null.

src/main/java/com/google/sps/user/repository/UserRepositoryFactory.java
```diff
@@ -0,0 +1,22 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/** Creates a UserRepository according to the requested type. */
+public class UserRepositoryFactory {
+    
+    public static UserRepository getUserRepository(RepositoryType repositoryType) {
+        if (repositoryType == null) {
+            throw new IllegalArgumentException("repositoryType can't be null!");
+        }
+        switch(repositoryType) {
```
tlheng: fix spaces

src/main/java/com/google/sps/user/User.java
```diff
@@ -0,0 +1,90 @@
...
+            this.imgUrl = imgUrl;
+            return this;
+        }
+
+        public Builder addSelfIntroduction(String selfIntroduction) {
+            this.selfIntroduction = selfIntroduction;
+            return this;
+        }
+
+        public Builder setPublicPortfolio() {
```
tlheng: I suggest to use a param to set. It is more readable and easier to use.

src/test/java/com/google/sps/DatastoreUserRepositoryTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private static final String ID_B = "useridB";
+
+    private static final String ID_C = "useridC";
+    private static final String EMAIL_C = "user@gmail.com_C";
+    private static final String NAME_C = "username_C";
+    private static final String SELF_INTRODUCTION_C = "I am the user_C";
+    private static final String IMG_URL_C = "/img.com_C";
+
+    private final static User toSaveUser = new User.Builder(ID, EMAIL).setName(NAME).addSelfIntroduction(SELF_INTRODUCTION).addImgUrl(IMG_URL).build();
```
tlheng: Here and below: Longer than 100 chars. You can break the line as follow:

```suggestion
    private final static User toSaveUser = 
        new User.Builder(ID, EMAIL)
            .setName(NAME)
            .addSelfIntroduction(SELF_INTRODUCTION)
            .addImgUrl(IMG_URL)
            .build()
```

tlheng (approved):

src/main/java/com/google/sps/user/repository/UserRepository.java
bori00: Done.

src/main/java/com/google/sps/user/repository/UserRepository.java
bori00: Added.

src/main/java/com/google/sps/user/repository/UserRepositoryFactory.java
bori00: Fixed.

src/main/java/com/google/sps/user/User.java
bori00: Added a parameter.

src/test/java/com/google/sps/DatastoreUserRepositoryTest.java
bori00: Broke the line. 
I set up an IntelliJ setting which does this for me automatically, and used the format suggested by IntelliJ.

PR #7, opened 2020-08-09
User Portfolio Form with Blobstore

bori00:
The recent changes enable users to save and update their portfolio data through a form. 

These changes consist of the following steps:
- when the user accesses the website for the very first time, their data(id and email address) is saved automatically in the database. It's the **NewUserServlet**'s responsibility to check if the currently logged in user exists in the database, and to save them if they don't. For this purpuse, I added the **existingUser()** function to UserRepository. 
- the user is able to save or update their current data in the database by submitting the portfolioForm. The user can also upload a profile photo. For this purpose, I fetched an upload-url generated by BlobstoreService, using **BlobstoreUploadUrlServlet,** which is set to be the action url of the portfolio form. When the user submits the form, the uploaded file is first processed by Blobstore, and only then is the request redirected to UserDataServlet.
- **UserDataServlet**  handles the update of the user's data, with the following modifications since the last state:
 -- the key of the uploaded image is received due to BlobstoreService, and it is stored together with the user's data
 -- the user's metadata is overwritten with the new data, but the photo is 
a) overwritten, if the user uploaded a new image
b) kept, if the user didn't upload a new image and didn't opt for deleting the old one either
c) deleted, if the user explicitly chose to do that. For this purpose a **new checkbox** is added to the form. 
 -- if the user's previous photo was overwritten/the user deleted the old photo, than the blob corresponding to it is deleted
- whenever the user accesses the page with the portfolio form, their currently saved data is fetched from the database(using **UserDataServlet**) and displayed in the form. This includes displaying the profile photo of the user, relying on **ServeBlobServlet**.

BlobstoreUploadUrlServlet and ServeBlobServlet are independent of the portfolio form. The form for which  the upload-url is requested can be specified though a parameter for BlobstoreUploadUrlServlet, and it can take any value from the **FormType** enum. The FormType enum is responsible for returning the corresponding form-handler servlet for the given form type. This way, these two servlets can be later reused for other forms as well.


src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -0,0 +1,153 @@
...
+   * some data saved) However, the user's photo is kept if they didn't submit a new one, unless the
+   * user specifically exressed their preference to drop the photo from their profile. Note:
+   * whenever this method is invoked in real-life, it's guaranteed that prevUserData is not null In
+   * case of tests, prevUserData is always null.
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    User user = getUserFromRequest(request);
+    User prevUserData = userRepository.getUser(userService.getCurrentUser().getUserId());
+    if (prevUserData != null && prevUserData.getImgKey() != null && !prevUserData.getImgKey().equals(user.getImgKey())) {
```
denniswillie: I think this line is a bit too long, maybe we can break it?

src/main/webapp/portfolioForm.js
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Fetches a url to which the files can be uploaded, relying on Blobstore API.
+ */
+function getBlobstoreUploadUrlFromServlet(formType) {
+  var url = new URL("/blobstore-upload-url", document.URL);
+  url.searchParams.append('formType', formType)
+  return fetch(url)
+      .catch(error => console.log("getBlobstoreUploadUrlFromServlet: failed to fetch: " + error))
+      .then(uploadUrl => uploadUrl.text())
+      .catch(error => console.log('getBlobstoreUploadUrlFromServlet: failed to convert to text: ' + error))
```
denniswillie: too long, maybe we can break it down?

denniswillie (commented):
LGTM!

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -0,0 +1,153 @@
...
+    blobInfoFactory = new BlobInfoFactory();
+    userRepository = UserRepositoryFactory.getUserRepository(RepositoryType.DATASTORE);
+    userService = UserServiceFactory.getUserService();
+  }
+
+  /** For testing purposes. */
+  public UserDataServlet(BlobstoreService blobstoreService, BlobInfoFactory blobInfoFactory) {
+    this.blobstoreService = blobstoreService;
+    this.blobInfoFactory = blobInfoFactory;
+    userRepository = UserRepositoryFactory.getUserRepository(RepositoryType.DATASTORE);
```
denniswillie: Why do we not do the initialization on our fields?

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -0,0 +1,153 @@
...
+    userService = UserServiceFactory.getUserService();
```
denniswillie: this too?

src/main/java/com/google/sps/servlets/UserDataServlet.java
bori00: Done.

src/main/webapp/portfolioForm.js
bori00: Done.

src/main/java/com/google/sps/servlets/UserDataServlet.java
bori00: Moved initialization.

src/main/java/com/google/sps/servlets/UserDataServlet.java
bori00: Moved.

src/main/java/com/google/sps/servlets/NewUserServlet.java
```diff
@@ -21,67 +21,49 @@
...
-@WebServlet("/user-data")
-public class UserServlet extends HttpServlet {
-  public static final String NAME_INPUT = "name";
-  public static final String PUBLIC_PORTFOLIO_INPUT = "publicPortfolio";
-  public static final String PUBLIC_PORTFOLIO_INPUT_PUBLIC_VALUE = "public";
-  public static final String SELF_INTRODUCTION_INPUT = "selfIntroduction";
-  public static final String IMG_URL_INPUT = "imgUrl";
-
+@WebServlet("/new-user-servlet")
+public class NewUserServlet extends HttpServlet {
```
tlheng: Maybe UserCreationServlet is better name

Maybe also explain that this is because we use Google account for login. The first time user logins our site, we create create and store the user in the our data

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -0,0 +1,161 @@
...
+
+  private final BlobstoreService blobstoreService;
+  private final BlobInfoFactory blobInfoFactory;
+
+  /**
+   * For production.
+   */
+  public UserDataServlet() {
+    blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+    blobInfoFactory = new BlobInfoFactory();
```
tlheng: this(BlobstoreServiceFactory.getBlobstoreService(), new BlobInfoFactory());

so we don't repeat code.

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * For testing purposes.
+   */
+  public UserDataServlet(BlobstoreService blobstoreService, BlobInfoFactory blobInfoFactory) {
+    this.blobstoreService = blobstoreService;
+    this.blobInfoFactory = blobInfoFactory;
+  }
+
+  /**
+   * Saves the recently submitted userdata(updates it if the user already has some data saved) in
```
tlheng: fix space

src/main/webapp/user.js
```diff
@@ -0,0 +1,60 @@
...
+ */
+function saveUserInDatabase() {
+  return fetch('/new-user-servlet', {method: 'POST'})
+      .catch(error => "saveUserInDatabase: failed to post new user: " + error);
+}
+
+/**
+ * Checks if the currently logged in user is already saved in the database.
+ */
+function userPresentInDatabase() {
```
tlheng: Do we really need to expose this to client? Would it be better if this check as well as the save-user are done silently by the servlet?

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -0,0 +1,161 @@
...
+   * the database. Note: the user's name, self-introduction and portfolio status will be rewritten
+   * with the new data, whatewer it is. (even if the new data is empty and previously the user had
+   * some data saved) However, the user's photo is kept if they didn't submit a new one, unless the
+   * user specifically exressed their preference to drop the photo from their profile. Note:
+   * whenever this method is invoked in real-life, it's guaranteed that prevUserData is not null In
+   * case of tests, prevUserData is always null.
```
tlheng: We don't need to mention test in doc. 

If you want a not null prevUser in test you can make userservice a param and mock it in test

tlheng (commented):
Looks good!

src/main/java/com/google/sps/servlets/NewUserServlet.java
bori00: Changed name and completed javadoc.

src/main/webapp/user.js
bori00: Moved check to the servlet.

src/main/java/com/google/sps/servlets/UserDataServlet.java
bori00: Done.

src/main/java/com/google/sps/servlets/UserDataServlet.java
bori00: Done.

src/main/java/com/google/sps/servlets/UserDataServlet.java
bori00: I removed the test-related javadoc. Since I can't properly test storing new blobs in Blobstore anyway, it wasn't needed to mock userservice.

src/main/webapp/setupIndex.js
```diff
@@ -10,7 +10,7 @@
  * They can only access the page after logging in.
  * Otherwise, the site will prompt the user to Google login page.
  *
- * It is checked if the user logs in for the first time to the website.
+ * It is checked(the check is done by the servlet) if the user logs in for the first time to the website.
```
tlheng: space after word `.. checked (the ...`

tlheng (dismissed):

src/main/webapp/setupIndex.js
bori00: Fixed.

tlheng (approved):

PR #14, opened 2020-08-14
Map: Places

bori00:
### Contains representation of place guides and location choosing functionality

Since the last PR, 
- I added two new pages, the **myPlaceGuides.html** and **createPlaceGuide.html**
- The pages **display the places** with different icons based on their type. The discover page displays the public place guides(orange icons), the myPlaceGuides and createPlaceGuide pages show the private(yellow icons) and public place guides of the currently logged din user(the place guides' and the user's data is hard-coded for now).
- The places have an infowindow, based on their type. It can be opened by clicking on the marker, and closed by clicking on the marker, on the close button or anywhere on the map.
- On the createPlaceGuide page, the user can **choose a location**(green icon) for the new place guide by
 -- using the searchbox
 -- clicking on a point on the maps
 -- moving the marker around(dragging)
 -- clicking on the marker of a POI(point of interest)
- After having chosen any location for the new place guide, a "Save location" button is enabled, and the user can **save the currently chosen location**. The saved location is displayed with a blue icon. 

Check out the current stage here: https://fbori-step-2020.appspot.com/index.html


src/main/webapp/maps/placeGuide.js
```diff
@@ -0,0 +1,47 @@
...
+ * The place, similarly to the Place class, can be specified either by coordinates or through a place ID.
+ */
+class PlaceGuide {
+  constructor(place, databaseId, description, audioKey, audioLength, imgKey, creatorId, creatorName) {
+    this._place = place;
+    this._databaseId = databaseId;
+    this._description = description;
+    this._audioKey = audioKey;
+    this._audioLength = audioLength;
+    this._imgKey = imgKey;
```
denniswillie: Can you change to imageKey to make it the similar with the java class? :)

denniswillie (commented):
LGTM!

src/main/webapp/maps/createPlaceGuide_createMap.js
```diff
@@ -0,0 +1,53 @@
...
+  searchBox.init();
+  enableChoosingLocationByClick(map, chosenLocation);
+
+  // Save location functionality.
+  var locationSaver = new LocationSaver(map, chosenLocation);
+  locationSaver.init();
+
+  // Display user's place guides functionality.
+  // The user's ID is hardcoded for now,
+  // but it will be fetched from the server in the future.
```
tlheng: We can put a tag for it so that we can search for these later. Example:

// TODO(fbori): Fetch user ID from server once the servlet is ready.

tlheng (dismissed):
I am not expert in js. Maybe wait for Chris to comment.

src/main/webapp/maps/placeGuide.js
bori00: Done.

src/main/webapp/maps/createPlaceGuide_createMap.js
bori00: Done.

src/main/webapp/maps/createPlaceGuide_createMap.js
```diff
@@ -0,0 +1,50 @@
...
+ * creation.
+ * Display place guides: displays the place guides created by the currently logged in user.
+ */
+function createMap() {
+  var myMapOptions = {
+    zoom: 2,
+    center: new google.maps.LatLng(0, 0),
+    mapTypeId: 'roadmap',
+    restriction: {
+      latLngBounds: {north: 85, south: -85, west: -180, east: 180},
```
cgollner, line 20: What are these coordinates? Maybe add a comment

src/main/webapp/maps/createPlaceGuide_createMap.js
```diff
@@ -0,0 +1,50 @@
...
+  chosenLocation.triggerChosenLocationChangeEvent = true;
+  document.getElementById("map")
+      .addEventListener(CHOSEN_LOCATION_CHANGE_EVENT, function () {
+        chosenLocation.visible = true;
+      });
+  var searchBox = new SearchBox(map, chosenLocation, "search-box");
+  searchBox.init();
+  enableChoosingLocationByClick(map, chosenLocation);
+
+  // Save location functionality.
```
cgollner: Instead of these section comments (Geolocation, Choose location, search box), can we extract the code of each section into their own methods? 
That way it prevents this method from growing too much, removes the need for the comments, and makes sure that the methods are not doing things at different levels of detail.

src/main/webapp/maps/discover_createMap.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * This function initialises the map and adds some functionalities to it:
+ * Geolocation: to track the user's current location, display it and center the map around it.
+ * Searching: lets the user search for places and center the map around it.
+ * Display place guides: displays all the public place guides.
+ */
+function createMap() {
```
cgollner, line 7: Seems like this function name is the same as in the previous file. Should we rename it to avoid name collisions?

src/main/webapp/maps/discover_createMap.js
```diff
@@ -0,0 +1,34 @@
...
+  var geolocator = new Geolocator(map);
+  geolocator.init();
+
+  // Searching functionality.
+  var searchResult =
+      Place.constructPlaceBasedOnCoordinates(map, 0, 0,
+          "Search Result", PlaceType.SEARCH_RESULT, true);
+  searchResult.visible = false;
+  var searchBox = new SearchBox(map, searchResult, "search-box");
+  searchBox.init();
```
cgollner: Same comment about extracting this into methods.

src/main/webapp/maps/geolocation.js
```diff
@@ -0,0 +1,152 @@
...
+
+  static convertCurrentLocationToLatLng(location) {
+    var locationLatLng = {
+      lat: location.coords.latitude,
+      lng: location.coords.longitude
+    };
+    return locationLatLng;
+  }
+
+  init() {
```
cgollner, line 55: Maybe add documentation explaining what init does and when it should be called.

src/main/webapp/maps/geolocation.js
```diff
@@ -0,0 +1,152 @@
...
+    this._geolocationControlDiv.addEventListener("click",
+        event => this.onGeolocationControlEvent());
+  }
+
+  onGeolocationControlEvent() {
+    if (!this._trackLocation) {
+      this.enableLocationTracking();
+    } else {
+      this.disableLocationTracking();
+    }
```
cgollner: Micro nit: Consider putting the "positive" condition first.

> If your conditional looks like “if not … else …” then reverse it to put the positive case first.

https://drive.google.com/file/d/0B-KFDiTrQe9lWTRvODVraGk1Xzg/preview

src/main/webapp/maps/myPlaceGuides_createMap.js
```diff
@@ -0,0 +1,34 @@
...
+  // Searching functionality.
+  var searchResult =
+      Place.constructPlaceBasedOnCoordinates(map, 0, 0,
+          "Search Result", PlaceType.SEARCH_RESULT, true);
+  searchResult.visible = false;
+  var searchBox = new SearchBox(map, searchResult, "search-box");
+  searchBox.init();
+
+  // Discover public place guides functionality.
+  PlaceDisplayer.displayPlaceGuidesOfUser(map);
```
cgollner: Same comments as on previous files.

cgollner (changes requested):

src/main/webapp/maps/createPlaceGuide_createMap.js
bori00: I added a comment with a brief explanation, but the actual reason for using these coordinates can be found [here](https://en.wikipedia.org/wiki/Mercator_projection#Truncation_and_aspect_ratio.)

src/main/webapp/maps/discover_createMap.js
bori00: Each html page has a separate script file with a `createMap() `function, including the required functionalities for that required page. 
No page will include two of these files, so we can't have collisions.

src/main/webapp/maps/geolocation.js
bori00: Done.

src/main/webapp/maps/myPlaceGuides_createMap.js
bori00: Done.

src/main/webapp/maps/geolocation.js
bori00: Done.

src/main/webapp/maps/discover_createMap.js
bori00: Done.

src/main/webapp/maps/createPlaceGuide_createMap.js
bori00: Done.

cgollner (approved):

PR #16, opened 2020-08-19
Dennis dev place guide html and bookmarking

denniswillie:
In this pull request, I added 2 features, which are place guide creation form and bookmarking feature for place guides.
1. **Place guide creation form**.
This enables user to preview images and audio files before submitting. I will also automatically preview image and audio if user loads an existing place guide.
2. **Bookmarking place guides**
I added the functionality to bookmark place guides, remove place guides from bookmark list, and also get all the bookmarked place guides bookmarked by a user.


src/main/webapp/createPlaceGuideForm.js
```diff
@@ -0,0 +1,140 @@
...
+}
+
+function activatePreviewFeature() {
+
+  const imageInput = document.getElementById("imageKey");
+  const imagePreview = document.getElementById("imagePreview");
+  // Previewing files uploaded by user.
+  imageInput.addEventListener("change", function() {
+    const file = this.files[0];
+    console.log(file);
```
bori00: Remove stetements used for debugging.

src/main/webapp/createPlaceGuideForm.js
```diff
@@ -0,0 +1,140 @@
...
+      const reader = new FileReader();
+
+      reader.addEventListener("load", function() {
+        audioPlayer.setAttribute('src', this.result);
+        audioPlayer.play();
+
+      });
+
+      reader.readAsDataURL(file);
+      console.log(reader);
```
bori00: Remove.

src/main/webapp/createPlaceGuideForm.js
```diff
@@ -0,0 +1,140 @@
...
+      });
+}
+
+/**
+ * Sets the destination url of the post method for the createPlaceGuide form to uploadUrl.
+ */
+function setFormActionUrl(uploadUrl, formId) {
+  var form = document.getElementById("formId");
+  form.action = uploadUrl;
+}
```
bori00: Let's not duplicate code. I know that previously I have included these functions in the js file related to the protfolio form, but as now we have multiple forms using them, I would advise to extract them in a separate file which can be reused multiple times without duplicating the code.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
```diff
@@ -94,6 +127,61 @@ private Entity createPlaceGuideEntity(PlaceGuide placeGuide) {
...
+      int bookmarkedIdsCopyIndex = 0;
+      while (bookmarkedIdsCopyIndex < bookmarkedIdsCopy.size()) {
+        long placeGuideId = bookmarkedIdsCopy.get(bookmarkedIdsCopyIndex);
+        Key placeGuideEntityKey = KeyFactory.createKey(ENTITY_KIND, placeGuideId);
+        try {
+          Entity placeGuideEntity = datastore.get(placeGuideEntityKey);
+          bookmarkedPlaceGuides.add(getPlaceGuideFromEntity(placeGuideEntity));
+          bookmarkedIdsCopyIndex++;
+        } catch(EntityNotFoundException err) {
+          System.out.println("PlaceGuide entity does not exist or has already been removed.");
```
bori00: Remove print statement.

src/main/java/com/google/sps/user/User.java
```diff
@@ -43,10 +48,12 @@
     @Nullable
     private String imgKey;
     private boolean publicPortfolio = false;
+    private final Set<Long> bookmarkedPlaceGuidesIds;
 
-    public Builder(String id, String email) {
+    public Builder(String id, String email, Set<Long> bookmarkedPlaceGuidesIds) {
```
bori00: Why do we need `bookmarkedPlaceGuideIds` in the constructor? Cannot we use an empty HashSet as default value and set the `bookmarkedPlaceGuideIds` with an optional setter if needed?

src/main/webapp/createPlaceGuide.html
```diff
@@ -0,0 +1,89 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset = "UTF-8">
+    <title>Create Place Guide</title>
+    <style>
+      #imageDiv {
```
bori00: Put css in separate file. I already had a style.css, you can use that for now.

src/main/webapp/createPlaceGuide.html
```diff
@@ -0,0 +1,89 @@
...
+        <div id = "imageDiv">
+        <br>
+        <img src = "" alt = "image preview" id = "imagePreview">
+        <br>
+        </div>
+        <label for = "deleteImage">Delete previous image</label>
+        <input type = "checkbox" name = "deleteImage" value = "delete" id = "deleteImage">
+        <br>
+        <label for = "placeId">place id : </label><br>
+        <input type = "text" name = "placeId" id = "placeId" readonly><br>
```
bori00: Why not make this one hidden too? The user doesn't need to know the ID of the place.
What we could do is to display the place's name, provided by the Places API.

src/main/webapp/createPlaceGuide.html
```diff
@@ -0,0 +1,89 @@
...
+        <br>
+        <label for = "description">Description:</label><br>
+        <textarea name = "description" rows = "4" cols = "50" id = "description"></textarea>
+        <br>
+        <label for = "length">Amount of time needed:</label><br>
+        <input type = "text" name = "length" id = "length"><br>
```
bori00: This should be a numeric value.

src/main/webapp/createPlaceGuide.html
```diff
@@ -0,0 +1,89 @@
...
+        </select>
+        <br>
+        
+        <fieldset>
+            <legend>Coordinate: </legend>
+            <label for = "latitude">latitude : </label><br>
+            <input type = "text" name = "latitude" id = "latitude" required><br>
+            <label for = "longitude">longitude : </label><br>
+            <input type = "text" name = "longitude" id = "longitude" required><br>
+            <br>
```
bori00: Do we want to allow the user to choose the position of the placeGuide by typing in the coordinates? If not, then these fields should be read-only.

src/main/webapp/createPlaceGuide.html
```diff
@@ -0,0 +1,89 @@
...
+        <label for = "length">Amount of time needed:</label><br>
+        <input type = "text" name = "length" id = "length"><br>
+        <label for = "isPublic">Place Guide Status:</label>
+        <select name = "isPublic" id = "isPublic">
+            <option value = "public">public</option>
+            <option value = "private">private</option>
+        </select>
+        <br>
+        
+        <fieldset>
```
bori00: The place(its name, based on my previous suggestion) also belongs logically to this fieldset, right? The user is basically interested in the location, which can be specified either by a place(with coordinates) or only by coordinates.

src/main/webapp/createPlaceGuideForm.js
```diff
@@ -0,0 +1,137 @@
...
+  if (placeGuides === undefined || placeGuides.length == 0) {
+    console.log("place guide does not exist yet.");
+  } else {
+    const placeGuide = placeGuides[0];
+    setFormInputValue(document.getElementById("id"), placeGuide.id);
+    setFormInputValue(document.getElementById("name"), placeGuide.name);
+    const audioSrc = new URL("/serve-blob", document.URL);
+    audioSrc.searchParams.append('blob-key', placeGuide.audioKey);
+    const audioPlayer = document.getElementById("audioPlayer");
+    audioPlayer.setAttribute("src", audioSrc);
```
bori00: Extract a function which finds an HTML element based on its ID, and sets its source to the url of the servlet with the given blobkey. 
Use it for the imagePreview as well.

src/main/webapp/createPlaceGuideForm.js
```diff
@@ -0,0 +1,137 @@
...
+    input.value = value;
+  }
+}
+
+function testGivenCoordinate() {
+  document.getElementById("latitude").setAttribute("value", 3.14);
+  document.getElementById("longitude").setAttribute("value", 2.56);
+}
+
+function fillFormWithPlaceGuideData(placeGuides) {
```
bori00: I understand that this function is for testing purposes, but still I think that logically it should only take one single placeGuide as argument, since that's the behavior that it will have in the future. Get the first element of the list before calling the function.

src/main/webapp/createPlaceGuideForm.js
```diff
@@ -0,0 +1,137 @@
...
+
+    if (file) {
+      const reader = new FileReader();
+      imagePreview.style.display = "block";
+
+      reader.addEventListener("load", function() {
+        imagePreview.setAttribute("src", this.result);
+      });
+
+      reader.readAsDataURL(file);
```
bori00: Can't we extract this part as a function and use it for the audioInput as well?

src/main/java/com/google/sps/placeGuide/repository/PlaceGuideRepository.java
```diff
@@ -3,12 +3,15 @@
...
 import com.google.sps.placeGuide.PlaceGuide;
 import java.util.List;
+import com.google.sps.user.User;
 
 /** Interface for handling place guide. */
 public interface PlaceGuideRepository {
 
   public void savePlaceGuide(PlaceGuide placeGuide);
 
+  public void bookmarkPlaceGuide(long placeGuideId, String userId);
```
bori00: Since bookmarking a placeGuide for a user means changing the user entity, not the placeGuide entity in the database, I would suggest putting this function in the UserRepository interface instead.

src/main/java/com/google/sps/placeGuide/repository/PlaceGuideRepository.java
```diff
@@ -23,5 +26,9 @@
   // (not available to be viewed by other users).
   public List<PlaceGuide> getCreatedPrivatePlaceGuides(String creatorId);
 
+  public List<PlaceGuide> getBookmarkedPlaceGuides(String userId);
+
+  public void removeBookmarkedPlaceGuide(long placeGuideId, String userId);
+
```
bori00: The previous remark is valid for this function as well. 
You may leave getBookmarkedPlaceGuides here, but I think it would fit into the UserRepository too.

src/main/java/com/google/sps/placeGuide/repository/PlaceGuideRepository.java
bori00: This approach would solve the problem of that `saveUpdatedUser` function, which really doesn't belong to the DatastorePlaceGuideRepository.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
```diff
@@ -94,6 +127,61 @@ private Entity createPlaceGuideEntity(PlaceGuide placeGuide) {
...
+      
+      // Update user with the new set.
+      saveUpdatedUser(user, bookmarkedPlaceGuidesIdsCopy);
+    } else {
+      throw new IllegalStateException("Cannot remove bookmarked place guides from non-existent user!");
+    }
+  }
+
+  private void saveUpdatedUser(User user, Set<Long> updatedBookmarkedPlaceGuidesIds) {
+    User updatedUser =
```
bori00: Instead of rebuilding the user object, we could implement a function for the User class to add a new placeGuideId to its list/remove one. Since you are using a set, this would be easier and also more efficient.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
```diff
@@ -52,6 +61,30 @@ private Entity createPlaceGuideEntity(PlaceGuide placeGuide) {
     return placeGuideEntity;
   }
 
+  @Override
+  public void bookmarkPlaceGuide(long placeGuideId, String userId) {
+    // Check if the corresponding placeGuide is in the database.
+    Key placeGuideEntityKey = KeyFactory.createKey(ENTITY_KIND, placeGuideId);
+    try {
+      Entity placeGuideEntity = datastore.get(placeGuideEntityKey);
```
bori00, line 73: As you don't actually need the placeGuideEntity, maybe you could implement a function which checks if a placeGudie exists or not in the database. Bookmark it only if it exists. 
It would be more readable then with these try-catch statements.

src/main/webapp/createPlaceGuideForm.js
denniswillie: done.

src/main/webapp/createPlaceGuideForm.js
denniswillie: done.

src/main/webapp/createPlaceGuideForm.js
denniswillie: done. I've created a script file called setUpFormBlobstore.js

src/main/webapp/createPlaceGuide.html
denniswillie: done.

src/main/webapp/createPlaceGuide.html
denniswillie: Hid placeId input tag.

src/main/webapp/createPlaceGuide.html
denniswillie: done.

src/main/webapp/createPlaceGuide.html
denniswillie: Since required and readonly cannot exist in the same tag, I have changed it with :
`required onkeypress="return false;"`
This way, the user still can "type" on the input form, but nothing will appear on the input form. The text cursor will also still appear when the user tries to type something on the input form. The input form will still be required.

src/main/java/com/google/sps/user/User.java
denniswillie: done.

src/main/java/com/google/sps/placeGuide/repository/PlaceGuideRepository.java
denniswillie: done.

src/main/java/com/google/sps/placeGuide/repository/PlaceGuideRepository.java
denniswillie: done, also moved the test methods used to test these functions from DatastorePlaceGuideRepositoryTest to DatastoreUserRepositoryTest.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
denniswillie: Staying with updating the user with an updated `bookmarkedPlaceGuidesIds` to keep it immutable.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
denniswillie: done.

src/main/webapp/createPlaceGuideForm.js
denniswillie: done.

src/main/webapp/createPlaceGuideForm.js
denniswillie: done.

src/main/webapp/createPlaceGuideForm.js
denniswillie: done.

src/main/webapp/createPlaceGuide.html
denniswillie: done.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
denniswillie: done.

src/main/java/com/google/sps/placeGuide/PlaceGuide.java
```diff
@@ -27,12 +27,13 @@
   private final long length;
 
   @Nullable
-  private final String description, imageKey;
+  private final String description, imageKey, placeName;
```
bori00: What is the placeName for?

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
bori00: Alright.

src/main/java/com/google/sps/servlets/PlaceGuideServlet.java
```diff
@@ -105,19 +125,26 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    String idStringValue = request.getParameter(ID_INPUT);
     long id;
+    String idStringValue = request.getParameter(ID_INPUT);
     if (!idStringValue.isEmpty()) {
       id = Long.parseLong(idStringValue);
     } else {
       // Create PlaceGuide entity id.
       Entity placeGuideEntity = new Entity(DatastorePlaceGuideRepository.ENTITY_KIND);
+      datastore.put(placeGuideEntity);
       id = placeGuideEntity.getKey().getId();
```
bori00: Can you move this part as a function to DatastorePlaceGuideRepository?

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
```diff
@@ -66,15 +73,88 @@ private User getUserFromUserEntity(Entity userEntity) {
...
+
+  private boolean placeGuideExists(Key placeGuideEntityKey) {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    try {
+      datastore.get(placeGuideEntityKey);
+      return true;
+    } catch(EntityNotFoundException err) {
+      return false;
+    }
+  }
```
bori00: This should be in the PlaceGuideRepository.

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
```diff
@@ -66,15 +73,88 @@ private User getUserFromUserEntity(Entity userEntity) {
...
+      bookmarkedPlaceGuidesIdsCopy.remove(placeGuideId);
+      
+      // Update user with the new set.
+      saveUpdatedUser(user, bookmarkedPlaceGuidesIdsCopy);
+    } else {
+      throw new IllegalStateException("Cannot remove bookmarked place guides from non-existent user!");
+    }
+  }
+
+  private void saveUpdatedUser(User user, Set<Long> updatedBookmarkedPlaceGuidesIds) {
```
bori00: Can we separate building the user and saving it to the database? Saving is already implemented, this function should handle building a new user with the updated bookmarkedPlaceGuideIds and returning the new User object only. Then you can save the the returned object, but these two tasks don't need to be done in the same function.

src/main/java/com/google/sps/servlets/PlaceGuideServlet.java
denniswillie: done, created a function with the name: saveAndGeneratePlaceGuideId.

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
denniswillie: done.

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
denniswillie: done. changed the name of the function to getUpdatedUser() and will return the updatedUser.

bori00 (commented):
LGTM!

pom.xml
```diff
@@ -100,7 +107,7 @@
                 <version>2.2.0</version>
                 <configuration>
                     <!-- TODO: set project ID. -->
-                    <deploy.projectId>fbori-step-2020</deploy.projectId>
+                    <deploy.projectId>denniswillie-step-2020</deploy.projectId>
```
cgollner: Can we avoid changing this and keep the real project id?

src/main/java/com/google/sps/placeGuide/repository/PlaceGuideRepository.java
```diff
@@ -3,12 +3,19 @@
...
 public interface PlaceGuideRepository {
 
+  public long saveAndGeneratePlaceGuideId();
+
   public void savePlaceGuide(PlaceGuide placeGuide);
 
+  @Nullable
+  public PlaceGuide getPlaceGuide(long placeGuideId);
+
   // @return all public {@code PlaceGuide}.
```
cgollner: Prefer javadoc style `/** javadoc */` for public methods, instead of `// code comments`. Same for the other methods.

src/main/java/com/google/sps/placeGuide/repository/PlaceGuideRepository.java
```diff
@@ -3,12 +3,19 @@
...
 
 /** Interface for handling place guide. */
 public interface PlaceGuideRepository {
 
+  public long saveAndGeneratePlaceGuideId();
+
   public void savePlaceGuide(PlaceGuide placeGuide);
 
+  @Nullable
+  public PlaceGuide getPlaceGuide(long placeGuideId);
```
cgollner: Adding `public` for methods is not needed, when the interface is public.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
```diff
@@ -94,6 +120,49 @@ private Entity createPlaceGuideEntity(PlaceGuide placeGuide) {
...
+        }
+      }
+      Set<Long> updatedBookmarkedIdsCopy = new HashSet<>(bookmarkedIdsCopy);
+      // Update and save user with updated {@code bookmarkedPlaceGuides}.
+      User updatedUser = getUpdatedUser(user, updatedBookmarkedIdsCopy);
+      userRepository.saveUser(updatedUser);
+      return bookmarkedPlaceGuides;
+    } else {
+      throw new IllegalStateException("Cannot get bookmarked place guides for non-existent user!");
+    }
```
cgollner: Prefer the "positive" condition first. In this case, if we do that, we can avoid the `else` and reduce nesting a little bit. 

```
if (user == null) {
 throw ...
}

// User related code
```

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
```diff
@@ -94,6 +120,49 @@ private Entity createPlaceGuideEntity(PlaceGuide placeGuide) {
...
+          bookmarkedPlaceGuides.add(getPlaceGuideFromEntity(placeGuideEntity));
+          bookmarkedIdsCopyIndex++;
+        } catch(EntityNotFoundException err) {
+          bookmarkedIdsCopy.remove(bookmarkedIdsCopyIndex);
+        }
+      }
+      Set<Long> updatedBookmarkedIdsCopy = new HashSet<>(bookmarkedIdsCopy);
+      // Update and save user with updated {@code bookmarkedPlaceGuides}.
+      User updatedUser = getUpdatedUser(user, updatedBookmarkedIdsCopy);
+      userRepository.saveUser(updatedUser);
```
cgollner: Why do we need to update bookmarked place guides, if this method is just about getting them?

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
```diff
@@ -94,6 +120,49 @@ private Entity createPlaceGuideEntity(PlaceGuide placeGuide) {
...
+
+  private User getUpdatedUser(User user, Set<Long> updatedBookmarkedPlaceGuidesIds) {
+    User updatedUser =
+        new User.Builder(user.getId(), user.getEmail())
+        .setBookmarkedPlaceGuidesIds(updatedBookmarkedPlaceGuidesIds)
+        .setName(user.getName())
+        .addSelfIntroduction(user.getSelfIntroduction())
+        .setPublicPortfolio(user.portfolioIsPublic())
+        .addImgKey(user.getImgKey())
+        .build();
```
cgollner: Would be nice if the `User` class had a `toBuilder()` method so you don't have to copy all these fields manually here.

src/main/java/com/google/sps/servlets/BookmarkPlaceGuideServlet.java
```diff
@@ -0,0 +1,72 @@
...
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    long placeGuideId = Long.parseLong(request.getParameter(PLACE_GUIDE_ID_PARAMETER));
+    String bookmarkHandlingType = request.getParameter(BOOKMARK_HANDLING_TYPE_PARAMETER);
+    BookmarkPlaceGuideQueryType queryType = 
+        BookmarkPlaceGuideQueryType.valueOf(bookmarkHandlingType);
+    handlePlaceGuide(queryType, placeGuideId);
+  }
+
+  private void handlePlaceGuide(
```
cgollner: Maybe rename to `togglePlaceGuideBookmark`? `handle` could be many things.

src/main/java/com/google/sps/servlets/PlaceGuideServlet.java
```diff
@@ -105,19 +124,23 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       id = Long.parseLong(idStringValue);
     } else {
-      // Create PlaceGuide entity id.
-      Entity placeGuideEntity = new Entity(DatastorePlaceGuideRepository.ENTITY_KIND);
-      id = placeGuideEntity.getKey().getId();
+      id = placeGuideRepository.saveAndGeneratePlaceGuideId();
+    }
+    String audioKey = getUploadedFileBlobKey(request, AUDIO_KEY_INPUT);
+    // The audioKey can be null if the user did not upload any audio file. Hence, it will
+    // get the previous audioKey.
```
cgollner, line 171: Nice usage of code comments to explain the "why" instead of the "what" 👍

src/main/java/com/google/sps/user/User.java
```diff
@@ -31,6 +34,9 @@
   @Nullable
   private final String imgKey;
 
+  // Contains the unique identifier of the corresponding bookmarked place guides.
```
cgollner: I think that there is no need for this comment. The field name is already self-explanatory.

src/main/java/com/google/sps/user/User.java
```diff
@@ -107,6 +120,10 @@ public boolean portfolioIsPublic() {
     return publicPortfolio;
   }
 
+  public Set<Long> getBookmarkedPlaceGuidesIds() {
+    return Collections.unmodifiableSet(bookmarkedPlaceGuidesIds);
+  }
```
cgollner, line 135: Nice that you are using an immutable version of the set 👍 
Would be nice if we could return ImmutableSet<Long> instead of Set<Long>. That way the users of this API know that they can't modify it. This doesn't have to be in this pull request though. 

https://guava.dev/releases/19.0/api/docs/com/google/common/collect/ImmutableSet.html

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
```diff
@@ -66,15 +76,83 @@ private User getUserFromUserEntity(Entity userEntity) {
...
+      Set<Long> bookmarkedPlaceGuidesIds = user.getBookmarkedPlaceGuidesIds();
+      Set<Long> bookmarkedPlaceGuidesIdsCopy = new HashSet<>(bookmarkedPlaceGuidesIds);
+      bookmarkedPlaceGuidesIdsCopy.remove(placeGuideId);
+      
+      // Update user with the new set.
+      User updatedUser = getUpdatedUser(user, bookmarkedPlaceGuidesIdsCopy);
+      saveUser(updatedUser);
+    } else {
+      throw new IllegalStateException("Cannot remove bookmarked place guides from non-existent user!");
+    }
```
cgollner: This method seems similar to the one above. Can we generalize them and have them call  a new private method `toggleBookmarkedPlaceGuide(long placeGuidId, String userId)` ? The method would check if the place is bookmarked, and change it to be the opposite?

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
```diff
@@ -66,15 +76,83 @@ private User getUserFromUserEntity(Entity userEntity) {
...
+
+  private User getUpdatedUser(User user, Set<Long> updatedBookmarkedPlaceGuidesIds) {
+    User updatedUser =
+        new User.Builder(user.getId(), user.getEmail())
+        .setBookmarkedPlaceGuidesIds(updatedBookmarkedPlaceGuidesIds)
+        .setName(user.getName())
+        .addSelfIntroduction(user.getSelfIntroduction())
+        .setPublicPortfolio(user.portfolioIsPublic())
+        .addImgKey(user.getImgKey())
+        .build();
```
cgollner: Again, if the `User` class had a `toBuilder()` method it would be nice :)

src/test/java/com/google/sps/PlaceGuideServletTest.java
```diff
@@ -0,0 +1,220 @@
...
+//   @Test
+//   public void doGet_queryBookmarkedPlaceGuides_returnCorrespondingPlaceGuides() {
+    
+//   }
+
+//   @After
+//   public void tearDown() {
+//     helper.tearDown();
+//   }
+// }
```
cgollner: Commented code?

cgollner (changes requested):

pom.xml
denniswillie: sure, it was to test the create place guide form.

src/main/java/com/google/sps/placeGuide/repository/PlaceGuideRepository.java
denniswillie: done.

src/main/java/com/google/sps/placeGuide/repository/PlaceGuideRepository.java
denniswillie: done.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
denniswillie: done.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
denniswillie: removed the updating of user's bookmarkedPlaceGuidesIds.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
denniswillie: implemented the toBuilder method.

src/main/java/com/google/sps/servlets/BookmarkPlaceGuideServlet.java
denniswillie: changed to togglePlaceGuideBookmark.

src/main/java/com/google/sps/servlets/PlaceGuideServlet.java
denniswillie: thanks!

src/main/java/com/google/sps/user/User.java
denniswillie: deleted the comment.

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
denniswillie: added the function with the suggested name.

src/main/java/com/google/sps/user/repository/impl/DatastoreUserRepository.java
denniswillie: done.

src/test/java/com/google/sps/PlaceGuideServletTest.java
denniswillie: I agreed with Bori that we'll write the test for this later, since both of us are working on the this same servlet. I removed this test.

cgollner (dismissed):

cgollner (dismissed):

tlheng (approved):

PR #19, opened 2020-08-24
Dennis dev place guide on list

denniswillie:
1. Create place guide div on list.
2. Add click event listener to the div so it highlights the chosen place guide div.
3. Add audio player.
4. Add DeletePlaceGuideButton
5. Add EditPlaceGuideButton
6. Add BookmarkPlaceGuideButton
7. Add DownloadPlaceGuide link.


src/main/webapp/PlaceGuideOnList.js
```diff
@@ -0,0 +1,41 @@
+class PlaceGuideOnList {
+  constructor(placeGuideId, place, name, creator, description, audioKey, audioLength, imageKey, createdByCurrentUser) {
+    const placeGuideDiv = document.createElement("div");
+    placeGuideDiv.setAttribute('id', placeGuideId);
```
bori00: Consider using an id which is specific for PlaceGuideOnLists only. For example "placeGuideOnList-{placeGuideId}"

src/main/webapp/PlaceGuideOnList.js
```diff
@@ -0,0 +1,41 @@
...
+    
+    placeGuideDiv.addEventListener('click', function() {
+      PlaceGuideManager.highlightPlaceGuide(placeGuideId);
```
bori00: highlightPlaceGuide is not a static function. 
I will create one placeGuideManager global variable for each page, please use that one.

src/main/webapp/PlaceGuideOnList.js
```diff
@@ -0,0 +1,41 @@
...
+    });
```
bori00: Please move this to a separate function, even if its short. You can use a naming pattern such as "highlightOnInfoboxClick", to make the code more readable than in case of using a listener directly in the constructor.

src/main/webapp/PlaceGuideOnList.js
```diff
@@ -0,0 +1,41 @@
...
+    createBlobView(audioKey, "audio", placeGuideDiv);
+    createBlobView(imageKey, "img", placeGuideDiv);
+
+    if (createdByCurrentUser) {
+      const deleteButton = document.createElement("button");
+      deleteButton.addEventListener("click", function() {
+        if (window.confirm("Click ok if you want to delete the place guide")) {
+          PlaceGuideManager.remove(placeGuideId);
+        }
+      });
```
bori00: Move this part to a separate function, such as "addDeleteButton".

src/main/webapp/PlaceGuideOnList.js
```diff
@@ -0,0 +1,41 @@
...
+
+      const editButton = document.createElement("button");
+      const queryString = 
+      editButton.addEventListener("click", function() {
+        window.location = "./createPlaceGuide.html?";
+      });
```
bori00: This one should also be a separate function. The constructor would be much easier to follow if you wouldn't have so many different variables, listeners, messages, etc, just a few function calls with suggestive names for what functionality a particular function is adding.

src/main/webapp/PlaceGuideOnList.js
```diff
@@ -0,0 +1,41 @@
...
+
+function createBlobView(blobKey, elementType, parentDiv) {
+  const element = document.createElement(elementType);
+  const src = new URL("/serve-blob", document.URL);
+  src.searchParams.append('blob-key', blobKey);
+  element.setAttribute("src", src);
+  parentDiv.appendChild(element);
+}
+
+function generateQueryString(placeGuideId, placeName, placeId, name, audioSrc, imageSrc, description, length, isPublic, latitude, longitude) {
```
bori00: What is this for?

src/main/webapp/PlaceGuideOnList.js
```diff
@@ -0,0 +1,41 @@
+class PlaceGuideOnList {
```
bori00, line 1: DownloadPlaceGuide and Bookmark/Unbookmark-PlaceGuide buttons are missing.

src/main/webapp/PlaceGuideOnList.js
```diff
@@ -0,0 +1,41 @@
...
+  constructor(placeGuideId, place, name, creator, description, audioKey, audioLength, imageKey, createdByCurrentUser) {
+    const placeGuideDiv = document.createElement("div");
```
bori00: Even if the design is not ready yet, at least add some text to the info-box containing the required information of the placeGuide(name, place, description, creator).

bori00 (commented):
I only need to check PlaceGuideOnList.js, right?
Please, in the future, when you start working on a new feature, which is independent of the ones currently being reviewed, start it from master, so that it doesn't contain any new, possibly unchecked code. Now your CL contains all the code from two other features(CLs), and it's more difficult to find what needs to be reviewed. Plus, it's risky that the merges won't be performed in the correct order, and then you may merge unchecked code. 
This applies only if the features from different CLs are not related to each other.

src/main/webapp/PlaceGuideOnList.js
```diff
@@ -0,0 +1,41 @@
...
+    if (createdByCurrentUser) {
+      const deleteButton = document.createElement("button");
+      deleteButton.addEventListener("click", function() {
+        if (window.confirm("Click ok if you want to delete the place guide")) {
+          PlaceGuideManager.remove(placeGuideId);
+        }
+      });
+
+      const editButton = document.createElement("button");
+      const queryString =
```
bori00: Is there something missing here?

src/main/webapp/PlaceGuideOnList.js
```diff
@@ -0,0 +1,41 @@
+class PlaceGuideOnList {
+  constructor(placeGuideId, place, name, creator, description, audioKey, audioLength, imageKey, createdByCurrentUser) {
```
bori00: We should add a bookmarkedByCurrentUser field to the constructor too. 
Sorry, I missed it when I created the diagram.

src/main/webapp/PlaceGuideOnList.js
denniswillie: changed to `placeGuideOnList-{placeGuideId}`

src/main/webapp/PlaceGuideOnList.js
denniswillie: changed to placeGuideManager.

src/main/webapp/PlaceGuideOnList.js
denniswillie: created a function called `highlightOnInfoboxClick`

PR #22, opened 2020-08-26
PlaceGuideInfo

bori00:
The recent changes extend the previous PlaceGudieWithCreator pair class with additional information about the placeGuide's relation to the currently logged in user. PlaceGuideWithCreator is renamed to PlaceGuideInfo.

- it is checked and stored whether the PlaceGuide **is created by the current user**
- it is checked and stored whether the PlaceGuide **is bookmarked by the current user**

This data is needed on the front-end side, and for an easier structuring of the information related to one PlaceGuide I included them in one single class.


denniswillie (commented):
LGTM!

tlheng (approved):

PR #24, opened 2020-08-26
StContainsFilter workaround

bori00:
#### Context
I previously implemented the map-related queries to the database with [StContainsFilter](https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/Query.StContainsFilter) and a [GeoRegion.Rectangle](https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/Query.GeoRegion.Rectangle). This filter checked if a particular placeGuide's coordinates are withing the bounds specified in form of a rectangle. 
These queries are needed because when the map's bounds are changed(by zooming/panning), we need to dynamically request the placeGuides in the new map area(with the condition that the zoom is higher then a threshold value).

#### Issue
Although the queries passed all the unit tests and the placeGuide requests were working properly on the devserver, the queries broke on the live server. The reason was that StContainsFilter is no longer supported by Datastore, and it only works on the test server. 
I got the exception: `java.lang.IllegalArgumentException: Geospatial Search has been turned down. StContainsFilter is no longer supported.`

#### Solution
As it turns out, Datastore currently doens't support any kind of geospatial searches, although it supports storing GeoPts for coordinates. The new approach relies on the comparator of GeoPt, which allows me to perform a query for getting the placeGuides whose latitude is within the southern and northern bounds. I get all these placeGuides, and then I need to filter them in-memory to remove the ones which are outside of the longitude bounds.
To avoid having too many placeGuides in memory, we will need to increase the threshold value for the zoom level above which the query is sent to the server.

Remark that for performing more complex queries I had to define [indexes](https://cloud.google.com/datastore/docs/concepts/indexes?_ga=2.259558666.-1423871165.1594566710&_gac=1.52765468.1597248050.EAIaIQobChMIgaylwYSW6wIVBgB7Ch2IdwgBEAAYASAAEgLLmPD_BwE) for the project. 

#### Other solutions considered
- Storing latitude and longitude in separate properties in the database, using inequality filters for both: doesn't work because Datastore allows inequality filters for one property only.
- Storing the latitudes and longitudes in two separate data tables(different entities) together with the corresponding placeGuide's ID, performing queries on both, getting the intersection of the ID sets(the intersection represents the set of IDs of placeGuides inside the map area) and then one-by-one get the placeGuides corresponding to these IDs: wouldn't decrease the amount of memory needed significantly, but would be much slower because of the high number of database accesses
- implementing my own StContainsFilter by extending the Filter class: it's impossible, because the Filter class is not extendable outside its package(it doesn't have a public constructor).


src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
```diff
@@ -117,53 +115,99 @@ public PlaceGuide getPlaceGuide(long placeGuideId) {
...
+    return placeGuidesCopy;
+  }
+
+  private boolean isBetweenLongitudeBounds(float longitude, float westBound, float eastBound) {
+    if (westBound < eastBound) {
+      return westBound <= longitude && longitude <= eastBound;
+    } else {
+      // The bounded area crosses the International Date Line(located at longitude 180/-180).
+      // In this case, the westBound is positive and the eastBound is negative.
+      if (longitude > 0) {
```
denniswillie, line 205: maybe change it to else if statement?

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
```diff
@@ -117,53 +115,99 @@ public PlaceGuide getPlaceGuide(long placeGuideId) {
...
+        return longitude >= westBound;
+      } else {
```
denniswillie: maybe delete the else statement?

denniswillie (commented):
LGTM!

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
bori00: Wouldn't do that. You are right that it would make the code shorter, but I think this structure offers a better readability. This way, it is clear that I want to handle separately the cases when the rectangle crosses/it doesn't cross the IDL.

src/main/java/com/google/sps/placeGuide/repository/impl/DatastorePlaceGuideRepository.java
bori00: Done.

cgollner (approved):

PR #27, opened 2020-08-28
Bug Fix: Multiple users mapped to same userId

bori00:
#### Issue
When multiple users, with different accounts logged in to the website and created PlaceGuides, all the PlaceGuides appeared to be created by the sameUser. It could be remarked, that the very first used who created a PlaceGuide showed up as the creator of all PlaceGuides. However, each user could see all PlaceGuides on the myPlaceGuides and the createPlaceGuide pages, as if they were their own. 

#### Explanation
The reason for this unexpected behavior was that the userId was requested only once from the UserService, when PlaceGudieServlet was constructed. Then, userId was saved as PlaceGuideServlet's instance variable. 
Given that HttpServlets are not guaranteed to be recreated each time a request is sent, the very first user's userId was kept in PlaceGuideServlet, even if new requests from other users were sent in the meantime. All PlaceGuides were saved in the database with this userId, and also, when a query for getting a user's PlaceGuides was sent, this userId was used in the query.

#### Solution. 
I modified PlaceGuideServlet so that it doesn't save userId in an instance variable, but it requests it from the UserService whenever a request is sent to it.
As the same issue would have arisen in case of BookmarkedPlaceGuideServlet(which wasn't tested yet on the live server), I modified that one too.


denniswillie (commented):
LGTM

tlheng (dismissed):

tlheng (approved):

PR #28, opened 2020-08-30
Forms : Resolve, File Preview, and Form Validation.

denniswillie:
In this pull request, I pushed changed related to the **CreatePlaceGuide** and **Portfolio** forms. I created this pull request since I did a mistake when I pushed the previous related pull requests, where I forgot to pull from master before pushing to this repo. Hence, there were too many changes to review. Below are the changes for this pull request:

1. **Resolve create place guide form bug**. This bug occurs when user submits a new place guide without attaching an image for the place guide and It throws **NullPointerException**. This happened, because whenever users create a new place guide, the `PlaceGuideServlet` will generate entity with a unique id for the particular place guide. Then the server will retrieve the entity and try to build a place guide (treats it as an existing place guide). The server needs an "existing place guide" for retrieving the "existing image file". The error occur because `PlaceGuide` has some non-nullable fields, and since the server builds a new place guide with no existing fields other than the id, this returns the error. I fixed it by changing when the server will build the already existing place guide, that is, when the submitted place guide data already has an id, in other words, the submitted data already has an existing place guide data in the database.

2. **File preview**. I enabled image previewing on index.html. Users can immediately see the image that was uploaded.

3. **Form Validation**. I implemented form validation for name, description, coordinate, and self introduction inputs, and give related validation message.


bori00 (commented):
LGTM!

cgollner (approved):

PR #29, opened 2020-08-31
MarkerClusterer: solve the issue of blinking markers

bori00:
#### Context
Whenever the map bounds are changed, on most of the pages the PlaceGuides in the newly defined map area are requested from the server, and they are displayed on the map with the help of markers.

#### Issue
When the new markers are displayed, some of them are blinking. Although this does not cause any bugs in the functionalities, it destroys user-friendliness as it is pretty annoying to see blinking markers whenever a map is moved a little. 

#### Solution
For the original implementation it has been taken into account that markers fall into 3 categories:

1. remaining markers(markers which were and remain on the map)
-.> no action need to be performed on these
2. markers which should be removed(they were on the map previously, but will be removed now)
-> They should be removed from the map and the dictionary of placeGuideOnMaps
3. markers which should be added(were not on the map previously, but will be from now on)
-> they must be created and added to the dictionary of PlaceGuideOnMaps

The problem was actually caused by not taking these categories into account when handling the MarkerClusterer. On each update, all markers where removed from the clusterer, and the all the new ones + remaining ones were added to it again. As the MarkerClusterer overwrites the visualisation of markers, even if a remaining marker wasn't explicitly removed and then added back to the map in my code, this was done by MarkerClusterer. 
To solve the issue, I updated the code related to MarkerClusterer so that markers are removed and added individually to it:

1. no action is taken on remaining markers. They remain in MarkerClusterer without any removal/addition.
2. only markers which should be removed are removed from MarkerClusterer too
3. only new markers are added to MarkerClusterer.

This approach solves the issue and makes the UI much more user-friendly and improves the efficiency of the code.


denniswillie (commented):
LGTM

cgollner (approved):

PR #30, opened 2020-08-31
Bookmarked PlaceGuides page

bori00:
#### This PR contains the changes needed to display a new page: the "bookmarked PlaceGuides" page. 

This new page is similar to the previous ones in that it contains a map showing some PlaceGuides. I set up an html page for it and javascript function to initialize it with the corresponding data. 
The differences are the following: 

- A **new query** to the database is supported: a query which gets the PlaceGuides bookmarked by the current user
- When this page will be displayed, since the user will have only a limited number of bookmarked PlaceGuides, **the maps bounds will be adjusted so that at the beginning all bookmarked PlaceGuides will be visible.** The displayed map area will be the **minimum one** which contains all of the bookmarked PlaceGuides. The reasons for this modified approach in case of this page are: 
 -- we expect the user to have bookmarked PlaceGuides close to each other, probably all of them being relevant for the user's next trip. By adjusting the map we help the user so that they are not required to search again for the location of the bookmarked PlaceGuides, but they can see them directly. On the other hand, the user will see all PlaceGuides at once on the list as well. 
-- given the limit of the bookmarked PlaceGuides, it's not problematic to hold all of the required data in-memory.


denniswillie (commented):
LGTM

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
```diff
@@ -6,12 +6,81 @@ class MapPlaceGuideDisplayer {
...
+    }
+  }
+
+  // This function scans all the placeGuides to find the minimum map area
+  // which contains all of them.
+  // To do so, it finds the minimum and maximum latitude coordinate of all
+  // placeGudies. These will be set as the southern and northern bound of the
+  // map area.
+  // For longitudes, it firs finds the maximum lane which doesn't contain any
+  // placeGuides(for this purpose, placeGuides are sorted based on their
```
cgollner: This comment doesn't seem to belong here?

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
```diff
@@ -6,12 +6,81 @@ class MapPlaceGuideDisplayer {
...
+  }
+
   update(placeGuides) {
     this.removePreviousPlaceGuidesFromMap(placeGuides);
     this.addNewPlaceGuidesToMap(placeGuides);
     this.updateMarkerClusters();
   }
 
+  // Sorts placeGuides in an ascending order based on their
+
```
cgollner, line 20: Remove extra line

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
```diff
@@ -6,12 +6,81 @@ class MapPlaceGuideDisplayer {
...
+      if (westLng > 180) westLng -= 360;
+      const eastLng = maxLngDifferenceWestCorner;
+      const southWestCorner = new google.maps.LatLng(minLat, westLng);
+      const northEastCorner = new google.maps.LatLng(maxLat, eastLng);
+      map.setZoom(15);
+      map.fitBounds(
+          new google.maps.LatLngBounds(southWestCorner,
+              northEastCorner),
+          10);
+    }
```
cgollner: This function is a little long. Can we try to break into multiple smaller ones for readability?

cgollner (changes requested):

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
bori00: You're right. Seems like one of the formatters that I applied changed the order of the comments and functions. 
I'm not sure about what happened, but I changed it manually.

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
bori00: Done.

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
bori00: After a quick check, it seems like it was caused by the formatted in IntelliJ. I will investigate this later.

src/main/webapp/pages/bookmarkedPlaceGuides_initPage.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * This function initialises the map and adds some functionalities to it:
+ * Geolocation: to track the user's current location, display it and center
+ * the map around it.
+ * Searching: lets the user search for places and center the map around it.
+ * Display place guides: displays the place guides created
+ * by the currently logged in user.
+ */
+let map;
```
widmoser, line 9: Where are these global variables used/read? I think we should try to not use global variables across files as this can lead to bugs that are hard to track, especially since we don't know which file depends on which.

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
bori00: Done.

src/main/webapp/pages/createPlaceGuide_initPage.js
```diff
@@ -11,27 +11,27 @@
  * by the currently logged in user.
  */
 
-var map;
-var placeGuideManager;
+let map;
```
widmoser, line 14: You could consider splitting up the change of adding the new page and refactoring the var/let/const declarations. It would make it a bit easier to review the change and lead to a bit clearer history in the repo.

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
widmoser: +1. I think we can try to simplify the logic as well. What about normalizing all the degree values so they are in the range [-180, 180] or some other canonical range. Then we get min and max as we do with lat. Finally we just need to check if (max - min) > 180 and if that is the case switch east and west. Maybe some of that is even already handled by the fitBounds method.

cgollner (dismissed):

src/main/webapp/pages/createPlaceGuide_initPage.js
bori00: You are right, sorry for that. 
The truth is that I have been using vars and eventually consts until the javascript linter was set up. Since then, I try to address the issues mentioned by the linter before merging the PR, which of course are related only to the lines which were changed in the current PR.
As you suggested, I will create a separate PR for revising the variable types in all files, in the older code too.

src/main/webapp/pages/bookmarkedPlaceGuides_initPage.js
bori00: Yes, that may indeed cause unexpected bugs. However, I decided to postpone the removal of these global variables because I am planning some major changes in the organisation of the pages' content, which will probably affect the scope of these variables too and will make the removal easier.

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
bori00: Update: we discussed that the solution mentioned above does not cover all the cases(for example, consider the PlaceGuides at longitudes -170, -160, 160, 170: the algorithm above would give the solution [170, -170], but this one doesn't contain all PlaceGuides. The correct solution would be [160, -160].
Instead, a new, simpler and less error-prone approach was implemented which relies on the extend() method of LatLngBounds, provided by the Maps API.

bori00: Remark: this solution(the one based on the extend() function) won't always lead to an optimal solution, in contrast with the original algorithm based on sorting.

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
```diff
@@ -13,6 +13,42 @@ class MapPlaceGuideDisplayer {
...
+    this._placeGuidesOnMap[placeGuideId].unhighlight();
+  }
+
+  // This function sets the map bound to the minimum one
+  // which contains all the PlaceGuides.
+  adjustMapToShowAll() {
+    let mapBounds = new google.maps.LatLngBounds();
+    let guidesExist = false;
+    for (const placeGuideId in this._placeGuidesOnMap) {
+      if (this._placeGuidesOnMap.hasOwnProperty(placeGuideId)) {
```
widmoser: What are the cases where placeGuideId appears in the for...in iteration but not as hasOwnProperty? If there is a difference you could consider using Object.getOwnPropertyNames to simplify this code a bit and avoid the "let" variable.

src/main/webapp/pages/bookmarkedPlaceGuides_initPage.js
widmoser: Ok SG.

src/main/webapp/pages/createPlaceGuide_initPage.js
widmoser: Ok great, thanks!

widmoser (dismissed):

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
bori00: Done.

cgollner (approved):

src/main/webapp/placeGuides/MapPlaceGuideDisplayer.js
```diff
@@ -13,6 +13,42 @@ class MapPlaceGuideDisplayer {
...
+    let guidesExist = false;
+    const placeGuideIds = Object.getOwnPropertyNames(this._placeGuidesOnMap);
+    placeGuideIds.forEach((placeGuideId) => {
+        const position =
+          this._placeGuidesOnMap[placeGuideId].marker.getPosition();
+        mapBounds.extend(position);
+        guidesExist = true;
+      }
+    );
+    if (guidesExist) {
```
widmoser, line 42: I think you can use "placeGuideIds.length > 0" here instead of managing the variable.

widmoser (approved):

PR #31, opened 2020-09-01
Location Choosing without Save Button

bori00:
The recent changes enable choosing the location of a PlaceGuide immediately by searching/dragging/clicking on pointor POI without having to click the "Save Location" button. 

The "Save Location" button and the corresponding class were deleted. The listeners are set to CHOSEN_LOCATION_CHANGE_EVENT instead. 
The scope of the change is the make the website more user-friendly.


denniswillie (commented):
LGTM

cgollner (approved):
Thanks for implementing this!

PR #32, opened 2020-09-01
Bug Fix: PlaceGuides with no placeId

bori00:
#### Issue
While testing I discovered that PlaceGuides which have coordinates only, but they are not linked to a place from google Places database(i.e. they don't have a placeId) cannot be displayed correctly.

#### Solution
The reason was that the form for the PlaceGuide's creation was submitted with incorrect placeId value: the placeId was simply set to "null"(as a string) if the picked location didn't have a placeId. I changed the form so that placeId is set to empty in this case, as this value is handled correctly by the servlet.


denniswillie (commented):
LGTM

cgollner (approved):

PR #34, opened 2020-09-07
Layout Changes

bori00:
#### General

- On the createPlaceGuide page, the form is now situated next to the map, on the left side.
- On the myportfolio page, the form is currently centered.

#### The menu
- The index page now contains the discover page, and the portfolio form is placed on a separate page. 
- The order of the pages has been changed.

#### The map
- The map is centered at GooglePlex.
- The threshold zoom level for sending a query to the server is increased to 8.

#### PlaceGuides' representation on the map
- Public PlaceGuides created by the current user are displayed with a lighter orange color, to indicate that the current user is the creator and they have the right to edit or delete it.

#### Map related actions

- The icons were changed to ones from Material Design.

#### Forms
I styled both the portfolio form and the placeguide form:
- I added a card containing all form elements to emphasize their connection
- I used text inputs, selection inputs(switches) and buttons from Material Design
- for the image preview feature, instead of the "delete previous image" checkbox the user has now a "clear" button with an "x" icon in the left corner of the image
- for the image preview feature, a default icon show up if no image is available
- for the publicity status of the portfolio/placeguide i replaced the dropdown with a switch
- in the PlaceGuide form the fieldset for the PlaceGuide's location is now hidden, but the infowindow of the marker showing the current location has been extended so that it contains all information about the chosen location. The location input is still required, but this is now enforced by enabling the submit button only after a location has been picked.

##### Portfolio Form
+ Remark the changed map action icons(enable geolocation and center map)
image

##### PlaceGuide Form
image


denniswillie (commented):
LGTM!

src/main/webapp/forms/formUtilities.js
```diff
@@ -0,0 +1,82 @@
+/**
+ * Sets one form input's value to the given value.
+ */
+function setFormInputValue(input, value) {
```
cgollner: Nit: maybe rename to `setFormInputValueOrEmpty`?

src/main/webapp/forms/formUtilities.js
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Whenever the clear-img-icon button is clicked, the current photo is removed,
+ * meaning that it won't be submitted
+ * (or, if the ucrrent photo is the previously saved photo,
+ * then it won't be kept)
+ * with the next submission.
+ * This change will be reflected in the preview too.
+ */
+function activateRemoveImageFeature(hasPreviousImage) {
```
cgollner: How about passing the element id as a parameter, like you did above for the files as well. That way this function would work with multiple elements.

src/main/webapp/forms/placeGuideForm.js
```diff
@@ -7,23 +7,54 @@ function setUpCreatePlaceGuideForm() {
   addBlobstoreUploadUrlToForm(
       'CREATE_PLACE_GUIDE_FORM', 'createPlaceGuideForm');
   activatePreviewFeature();
+  activateRemoveImageFeature(false);
+  styleInputs();
 }
 
+/**
+ * This function initialises the comonents managed by the Material Design library.
```
cgollner: typo: comonents

src/main/webapp/forms/placeGuideForm.js
```diff
@@ -7,23 +7,54 @@ function setUpCreatePlaceGuideForm() {
...
+  const chooseAudioFileButtonRipple = new mdc.ripple.MDCRipple(
+      document.getElementById("chooseAudioFileBtn"));
+  const chooseImageFileButtonRipple = new mdc.ripple.MDCRipple(
+      document.getElementById("chooseImageFileBtn"));
+  const deletePrevImageCheckbox = new mdc.checkbox.MDCCheckbox(
+      document.getElementById('deletePrevImageCheckbox'));
+  const deletePrevImageFormField = new mdc.formField.MDCFormField(
+      document.getElementById('deletePrevImageFormField'));
+  deletePrevImageFormField.input = deletePrevImageCheckbox;
+  const publicitySwitchControl = new mdc.switchControl.MDCSwitch(
```
cgollner, line 36: We are creating the variables, but they are never used?

src/main/webapp/maps/Place.js
```diff
@@ -158,7 +161,14 @@ class Place {
...
   getInfoWindowContent() {
-    const content = '<h3>' + this._name + '</h3>';
+    let content = '<h3>' + this._name + '</h3>';
+    if (this._mapsPlace != null) {
+        content += '<h4> Place: ' + this._mapsPlace.name + '</h4>';
+    }
+    // ToUrlValue is favored instead of toString because it has a parameter
+    // to set the precision.
+    content+=
+        '<p>' + 'Coordinates: (' + this._position.toUrlValue(3) + ')' +'</p>';
```
cgollner: Can we use string interpolation here instead of concatenation? (Not sure if Javascript supports this, but I think it might)

src/main/webapp/placeGuides/PlaceGuideOnMap.js
```diff
@@ -65,10 +65,12 @@ class PlaceGuideOnMap {
     if (creatorName == undefined) {
       creatorName = creator.email;
     }
-    const content = '<h3>' + name + '</h3>' +
+    let content = '<h3>' + name + '</h3>' +
         '<h4> Created by: ' + creatorName + '</h4>' +
-        '<h4> Place: ' + placeName + '</h4>' +
-        '<p>' + description + '</p>';
+        '<h4> Place: ' + placeName + '</h4>';
```
cgollner: Same question about maybe using String interpolation here. Would make this a bit easier to read.

cgollner (changes requested):

src/main/webapp/forms/placeGuideForm.js
bori00: Yes. Material design initializes the animations and some layout elements(for example moving the labels to up when a value is present for the textfields, character counter for the textareas, animations for the buttons and for the switches, etc.) required for these components after the initialization of the corresponding javascript objects.

src/main/webapp/forms/formUtilities.js
bori00: Done.

src/main/webapp/forms/placeGuideForm.js
bori00: Fixed.

src/main/webapp/forms/formUtilities.js
bori00: Done.

src/main/webapp/maps/Place.js
bori00: I used template literals.

src/main/webapp/placeGuides/PlaceGuideOnMap.js
bori00: Done.

cgollner (approved):

PR #35, opened 2020-09-07
Restructure PlaceGuideManager's data and responsibilities

bori00:
#### Reasons

Most of these changes became necessary because we added "Bookmarked PlaceGuides" page to the website. This page requires a slightly different approach for getting and displaying the placeguides, given that

- it displays all bookmarked placeguides at once, and their data must be kept in memory
- when a placeguide is unbookmarked on the "bookmarked placeguides" page, it must be immediately removed

Moreover, the ListPlaceGuideDisplayer will need some page-dependent data in the future. 

In addition to these, I also took into account that in the future the website should be converted to a single-page application. These changes should make the conversion easier, as only PlaceGuideManager is now page-dependent.

#### Changes
- PlaceGuideManager is responsible for knowing which page we are currently on, and adjust the commands sent to  the "executers" after taking into account the page. 
- PlaceGuideRepository no longer stores the QueryType. Instead, PlaceGudieManager sends it the query type each time the placeguides need to be updated. Thus, PlaceGuideRepository became independent from the current page.
- PlaceGuideManager is listening to changes in the map bounds, instead of letting the "update" function being called from the outside.
- I implemented toggleBookmark() for PlaceGuideManager to bookmark/unbookmark a placeGuide. If we are currently on "Bookmarked PlaceGuides" page, and the placeguide was unbookmarked, it will also be removed from the map.


src/main/webapp/placeGuides/PlaceGuideManager.js
```diff
@@ -1,32 +1,61 @@
...
+      query: PlaceGuideRepository.QUERY_TYPE.BOOKMARKED,
+    }
+  };
+
+  constructor(page, map) {
+    this._page = page;
+    this._placeGuideRepository = new PlaceGuideRepository();
     this._highlightedPlaceGuideId = null;
     this._mapPlaceGuideDisplayer = new MapPlaceGuideDisplayer();
     this._listPlaceGuideDisplayer = new ListPlaceGuideDisplayer();
```
denniswillie, line 33: so we should pass the query type as an argument for the constructor right?

bori00: Yes, we will, as soon as we'll have the updated ListPlaceGuideDisplayer.

src/main/webapp/placeGuides/PlaceGuideManager.js
```diff
@@ -1,32 +1,61 @@
 /**
  * This class is responsible for connecting all the placeGuide-related
  * objects: the two displayers and the repository.
- * When an event occurs, PlaceGuideManager is triggered and it
- * sends the commands to the
- * classes which are responsible for handling the events.
+ * When the map's bounds are changed, PlaceGuideManager
+ * sends the eneded commands to all classes involved in
```
cgollner: typo: eneded

cgollner (changes requested):

src/main/webapp/placeGuides/PlaceGuideManager.js
```diff
@@ -1,32 +1,61 @@
...
+      google.maps.event.addListener(map, 'idle', function () {
+        thisManager.update(map.getBounds(), map.getZoom());
+      });
+    }
   }
 
-  update(bounds, zoom, showAll) {
-    this._placeGuideRepository.updatePlaceGuides(bounds, zoom)
+  update(bounds, zoom) {
+    this._placeGuideRepository.updatePlaceGuides(this._page.query, bounds, zoom)
```
widmoser: "Update" can be ambiguous here because it could also mean that place guides (on the server) are updated. Maybe this can be called refreshPlaceGuides or fetchPlaceGuides.

src/main/webapp/placeGuides/PlaceGuideManager.js
```diff
@@ -1,32 +1,61 @@
...
         .then((response) => {
           const placeGuides = this._placeGuideRepository.placeGuides;
           this._mapPlaceGuideDisplayer.update(placeGuides);
           this._mapPlaceGuideDisplayer.update(placeGuides);
-          if (showAll) {
-              this._mapPlaceGuideDisplayer.adjustMapToShowAll();
+          if (this._page === PlaceGuideManager.PAGE.BOOKMARKED_PLACEGUIDES) {
```
widmoser, line 51: I think it would be better to put this as a field in the page enumeration similar to the query and have a generic implementation here as before.

src/main/webapp/placeGuides/PlaceGuideManager.js
```diff
@@ -45,4 +74,12 @@ class PlaceGuideManager {
     this._listPlaceGuideDisplayer.unhighlight(this._highlightedPlaceGuideId);
     this._highlightedPlaceGuideId = null;
   }
+
+  toggleBookmark(placeGuideId) {
```
widmoser, line 86: Having this here is not ideal as it adds conditional code based on the page. If it's only one its not so bad, but imagine you continue working on this and more and more features are added over time resulting in a complex if...else logic tree that doesn't encapsulate logic and will be very hard to understand.

It would be good to move page specific logic into a page adapter or have a more generic implementation (e.g. toggling a bookmark forces a refresh on the guides from the server).

Also note that for optimistic updates you need to handle error cases or the client cache might get out of sync with the server (i.e. on the client the bookmark is removed but not on the server). In general it is a more simple approach to have a responsive server and wait for the operation to be applied.

src/main/webapp/placeGuides/PlaceGuideRepository.js
```diff
@@ -142,18 +146,23 @@ class PlaceGuideRepository {
 
   togglePlaceGuideBookmarkStatus(placeGuideId) {
     // Toggle in in-memory dictionary.
-    var isBookmarked = this._placeGuides.bookmarkedByCurrentUser;
-    this._placeGuides.bookmarkedByCurrentUser = !isBookmarked;
+    var isBookmarked = this._placeGuides[placeGuideId].bookmarkedByCurrentUser;
```
widmoser: const

src/main/webapp/placeGuides/PlaceGuideRepository.js
```diff
@@ -142,18 +146,23 @@ class PlaceGuideRepository {
...
+    this._placeGuides[placeGuideId].bookmarkedByCurrentUser = !isBookmarked;
     // Toogle in database.
     var url = new URL("bookmark-place-guide", document.URL);
```
widmoser: const

src/main/webapp/placeGuides/PlaceGuideRepository.js
```diff
@@ -142,18 +146,23 @@ class PlaceGuideRepository {
...
+    if (isBookmarked) {
       url.searchParams.append("bookmarkHandlingType", "REMOVE");
+    } else {
+      url.searchParams.append("bookmarkHandlingType", "BOOKMARK");
     }
     return fetch(url)
-        .catch(error => console.log("BookmarkPlaceGuideServlet: failed to fetch: "
-            + error));
+        .catch(error =>
+            console.log("BookmarkPlaceGuideServlet: failed to fetch: "
```
widmoser: You should show an error message to the user as well (e.g. a toast) and in case of optimistic update revert the local cache.

src/main/webapp/placeGuides/PlaceGuideManager.js
bori00: Done.

src/main/webapp/placeGuides/PlaceGuideManager.js
bori00: Renamed to refrechPlaceGuides in PlaceGuideManager and fetchPlaceGuides in PlaceGuideRepository.

src/main/webapp/placeGuides/PlaceGuideRepository.js
bori00: Done.

src/main/webapp/placeGuides/PlaceGuideRepository.js
bori00: Done.

src/main/webapp/placeGuides/PlaceGuideManager.js
bori00: Done.

src/main/webapp/placeGuides/PlaceGuideManager.js
bori00: I converted it to a responsive server.

src/main/webapp/placeGuides/PlaceGuideRepository.js
bori00: I added error messages and used the responsive server concept, so that I update the local cache only after the fetch was successfully executed.

cgollner (dismissed):

widmoser (dismissed):

cgollner (approved):

PR #37, opened 2020-09-07
Centering the Map at Current Location

bori00:
This PR enables centering the map, each time a page is loaded, at the users current location, if it can be found.
In addition to that, a higher zoom level will be set, to provide a better overview of the user's area.


denniswillie (commented):
LGTM!

tlheng (approved):

PR #41, opened 2020-09-10
Limit Number of Bookmarked PlaceGuides

bori00:
### This PR contains the changes needed to limit the number of bookmarked placeguides to 25/user.

To achieve this goal, I applied the following changes:

#### Server Side 
- the User class has a field specifying the maximum number of bookmarked placeguides(currently set to 25) and a public function which specifies if a concrete user can bookmark one more placeguide
- The BookmarkPlaceGuideServlet checks in case of a bookmarking action if the current user can bookmark a new guide. The action is executed only if they can. The servlet returns true if the action was executed, and false, if it was not allowed.

#### Client Side
- the PlaceGuideRepository updates the local cache only if the servlet returned a positive result(i.e. the action was allowed)
- PlaceGuideRepository returns to placeGuideManager the status of the action, which can be success, not allowed or failed(server failure).
- PlaceGuideManager handles the status returned by PlaceGuideRepository: if the action completed successfully, it sends the necessary commands to the displayers to update the UI. Otherwise, a message will be displayed to the user.
- Only if the action was succesfully completed will the bookmark icon be changed. 

#### Tests
the tests of BookmarkPlaceGudieServlet were missing completely, so I implemented them all, taking into account the new changed.

### DIFF
--- a/src/main/java/com/google/sps/servlets/BookmarkPlaceGuideServlet.java
+++ b/src/main/java/com/google/sps/servlets/BookmarkPlaceGuideServlet.java
@@ -3,6 +3,7 @@ package com.google.sps.servlets;
 import com.google.appengine.api.datastore.*;
 import com.google.appengine.api.users.UserServiceFactory;
 import com.google.sps.data.RepositoryType;
+import com.google.sps.user.User;
 import com.google.sps.user.repository.UserRepository;
 import com.google.sps.user.repository.UserRepositoryFactory;
 import java.io.IOException;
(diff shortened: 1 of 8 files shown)


bori00:
Remark that this PR is converted to a draft because some further changed will be needed in ListPlaceGuideDisplayer and PlaceGuideOnList classes, but those are not ready yet.

denniswillie (commented):
LGTM!

src/test/java/com/google/sps/BookmarkPlaceGuideServletTest.java
```diff
@@ -0,0 +1,304 @@
...
+  // Users data.
+  private static final String ID_A = "useridA";
+  private static final String ID_B = "useridB";
+  private static final String ID_C = "useridC";
+  private static final String EMAIL = "user@gmail.com";
+  private static final Set<Long> EMPTY_BOOKMARKED_PLACE_GUIDES_IDS = null;
+  private static final Set<Long> BOOKMARKED_PLACE_GUIDES_IDS_A =
+      new HashSet<>(
+          Arrays.asList(
+              (long) 1, (long) 2, (long) 3, (long) 4, (long) 5, (long) 6, (long) 7, (long) 8,
```
tlheng: 1l instead of (long) 1

src/main/java/com/google/sps/servlets/BookmarkPlaceGuideServlet.java
```diff
@@ -29,27 +31,57 @@
...
       BookmarkPlaceGuideQueryType bookmarkPlaceGuideQueryType, long placeGuideId) {
     String userId = UserServiceFactory.getUserService().getCurrentUser().getUserId();
     switch (bookmarkPlaceGuideQueryType) {
       case BOOKMARK:
-        userRepository.bookmarkPlaceGuide(placeGuideId, userId);
-        break;
+        User user = userRepository.getUser(userId);
+        if (user == null) {
+          throw new IllegalStateException("The current user does not exist in the database!");
+        } else {
```
tlheng, line 70: no need for this else here

tlheng (approved):

src/test/java/com/google/sps/BookmarkPlaceGuideServletTest.java
bori00: Done.

src/main/java/com/google/sps/servlets/BookmarkPlaceGuideServlet.java
bori00: Removed.

PR #43, opened 2020-09-10
Geolocation Timeout

bori00:
#### Context
When the map is loaded on one page, I use navigator.geolocation to find the user's current location and center the map at that point. 

#### Issue
By default, the naviator.geoloctaion has an infinite timeout, so whenever the user's location is finally found, no matter how late it is, the map gets recentered. This may be annoying if the user already started taking actions(for example, if they already centered a map at another location in which they are interested).

#### Fix
I set a timeout so that the map gets recentered after the page is loaded only if the location is found within 1s.

### DIFF
--- a/src/main/webapp/maps/Geolocator.js
+++ b/src/main/webapp/maps/Geolocator.js
@@ -68,7 +68,8 @@ class Geolocator {
 
   /**
    * This function tries to find the user's current location, and
-   * if it is found, it centers the map around it and sets 
+   * if it is found within 1 second,
+   * it centers the map around it and sets 
    * a higher zoom level.
    * If the location cannot be found, nothing happens.
    */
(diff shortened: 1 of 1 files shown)


denniswillie (commented):
LGTM!

cgollner (commented):
Imagine the user is patiently waiting for the location to update, but then it never does because it timed out after a second.
Would it be possible to do it based on the user having interacted with the map yet? I.e. if the user moved the map, cancel location updates?

cgollner (approved):

bori00:
The user won't expect the map to be centered at their location when the page is loaded. I expect the user to start processing what they see after ~1s, so after that moment, it's confusing to see an unexpected change on the page. 
Remark that the user has an option to explicitly ask for the map to be centered at the current location, whenever they want.

bori00:
Further options were discussed offline.

PR #44, opened 2020-09-10
Saving user's data on every page

bori00:
I modified the initPage() function of every page so that whenever a user accesses a page, if their data is not saved yet in the database, then it will be saved now before accessing any features of the website.

The reason for this is that all the servlets assume that the current user has some data in the database, so users shouldn't be able to access the website without being present in the database. Otherwise, the website would crash.


denniswillie (commented):
LGTM!

src/main/webapp/pages/bookmarkedPlaceGuides_initPage.js
```diff
@@ -14,13 +14,15 @@ function initPage() {
...
       location.replace(userAuthenticationStatus.loginUrl);
     } else {
-      const menu = new Menu(Menu.PAGE_NAMES.BOOKMARKED_PLACEGUIDES);
-      const mapWidget = new MapWidget();
-      mapWidget.addGeolocationFunctionality();
-      mapWidget.addSearchingFunctionality();
-      map = mapWidget.map;
-      placeGuideManager = new PlaceGuideManager(
-          PlaceGuideManager.PAGE.BOOKMARKED_PLACEGUIDES, map);
+      saveUserInDatabase().then((response) => {
```
cgollner, line 17: Isn't this always saving it even if it is already saved?

bori00: No. The check for the user's existence in the database was previously moved to the server-side.

cgollner: But it still does a network request every time before proceeding, right? Could this delay be avoided or is it necessary?

bori00: It is necessary due to the way the servlets are organised: it has been always assumed that the current user exists in the database. 
This could be avoided in two ways: adding additional checks and handling the case of the user not being present in the database separately, or by creating a single-page application(which would reduce the number of such network requests to 1/session). 
However, given the time constraint that I have for the project, this is the only feasible solution at the moment.

cgollner (approved):

PR #48, opened 2020-09-15
Bug Fix: unhighlighting inexistent placeguides

bori00:
#### Context
When a user tries to highlight a placeguide, any previously highlighted guide will be "unhighlighted".

#### Issue
If previously highlighted guide is not on the current map area anymore, the corresponding infobox div and PlaceGuideOnMap object do not exist anymore. Any actions that should be performed on these object will cause a crash. 

#### Solution
To avoid this crash, I simply need to check if the corresponding objects still exist.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -109,7 +109,6 @@
                     <!-- TODO: set project ID. -->
                     <deploy.projectId>denniswillie-step-2020</deploy.projectId>
                     <deploy.version>1</deploy.version>
-                    <port>8181</port>
                 </configuration>
             </plugin>
           
--- a/pom.xml
+++ b/pom.xml
@@ -109,6 +109,7 @@
                     <!-- TODO: set project ID. -->
                     <deploy.projectId>denniswillie-step-2020</deploy.projectId>
                     <deploy.version>1</deploy.version>
+                    <port>8181</port>
                 </configuration>
             </plugin>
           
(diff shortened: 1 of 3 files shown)


bori00:
Please remark that some removals in the code refer to some unnecessary duplications in the code which may have been caused by a poorly resolved merge conflict.

pom.xml
```diff
@@ -109,6 +109,7 @@
                     <!-- TODO: set project ID. -->
                     <deploy.projectId>denniswillie-step-2020</deploy.projectId>
                     <deploy.version>1</deploy.version>
+                    <port>8181</port>
```
cgollner: Seems like this change is unrelated. Can we leave it out?

src/main/webapp/placeGuides/PlaceGuideOnList.js
```diff
@@ -418,8 +418,12 @@ class PlaceGuideOnList {
   static close(placeGuideId) {
     const divId = 'placeGuideOnList-' + '{' + placeGuideId + '}';
     const placeGuideDiv = document.getElementById(divId);
-    placeGuideDiv.querySelectorAll('.folded-placeGuide')[0].style.display = 'block';
-    placeGuideDiv.querySelectorAll('.card-placeGuide')[0].style.display = 'none';
-    placeGuideDiv.style.removeProperty('padding');
+    if (placeGuideDiv != null) {
```
cgollner: Should we compare to `undefined` instead? Or is it intentional?

cgollner (changes requested):

src/main/webapp/placeGuides/PlaceGuideOnList.js
bori00: It was intentional, because document.getElementById() returns null if the element is not found.

pom.xml
bori00: Sure. I removed it.

cgollner (approved):

PR #50, opened 2020-09-16
User's portfolio page

bori00:
##### PlaceGuide Info Box

- In the info box of any guide, a new icon appears, representing the creator of that particular guide. 
    - The icon consists of the creator's profile picture, if it exists, otherwise a default icon is used. 
    - If the user hovers on the icon, the name of the creator shows up, if available. Otherwise, the email address's first part(the one before the @) will appear
    - If the user clicks on the icon, and the portfolio of that user is public, then it will lead to the portfolio page of the creator. Otherwise a message will show up

Screenshot from 2020-09-16 11-37-05

##### Portfolio Page

- The portfolio page of user A consists of the following elements
    - a map showing the public guides of user A in the selected map area
    - a list showing the same guides
    - an info box containing the photo, name(or email) and self-introduction of user A

Screenshot from 2020-09-16 11-42-14

#### Changes in the server-side code
In PlaceGuideServlet a new query type is supported: CREATED_BY_GIVEN_USER_PUBLIC_IN_MAP_AREA, which requires as parameters the map bounds and the id of the given user. This relies on .getCreatedPublicPlaceGuidesInMapArea() database query. 

#### Changes in the client-side code
- UserRepresentation class: handles the display of the user icon in the placeguide info boxes, as described above
- usersPortfolio.html contains the new page, and usersPortfolio_initPage() handles its setup. Whenever the page is accessed, it receives through url parameters the data of the user, including the ID. The info box with the user data is filled based on the information sent through url parameters, whereas the  placeguides belonging to this user are fetched from the server base don the user ID.


pom.xml
```diff
@@ -109,6 +109,7 @@
                     <!-- TODO: set project ID. -->
                     <deploy.projectId>denniswillie-step-2020</deploy.projectId>
                     <deploy.version>1</deploy.version>
+                    <port>8181</port>
```
cgollner: Can this be done as a separate PR?

src/main/java/com/google/sps/data/PlaceGuideQueryType.java
```diff
@@ -16,23 +16,30 @@
...
+  ALL_PUBLIC(false, false),
+  CREATED_ALL(false, false),
+  CREATED_PUBLIC(false, false),
+  CREATED_PRIVATE(false, false),
+  BOOKMARKED(false, false),
+  ALL_PUBLIC_IN_MAP_AREA(true, false),
+  CREATED_ALL_IN_MAP_AREA(true, false),
+  CREATED_PUBLIC_IN_MAP_AREA(true, false),
+  CREATED_PRIVATE_IN_MAP_AREA(true, false),
+  CREATED_BY_GIVEN_USER_PUBLIC_IN_MAP_AREA(true, true);
```
cgollner: In certain contexts literal arguments - such as 0, "", true and false, or null - can make it difficult for readers to know what a method will do. When you must call such a method you're encouraged to use the parameter name as an inline comment at the call site, so that readers don't need to look at the method declaration to understand the parameter's purpose.

Such a comment should be formatted in this way:
```
booleanMethod(/* enableFoo= */ true);
varargsMethod(/* foos...= */ foo1, foo2);
```

src/main/webapp/pages/menu.js
```diff
@@ -43,12 +43,15 @@ class Menu {
   constructor(pageName) {
     const NO_TABS = 5;
     for (var i = 0; i < NO_TABS; i++) {
-      var tab = this.createMenuTab(i, i == pageName.index);
+      const isFocused = pageName != undefined && pageName.index == i;
```
cgollner: I think it is good practice to always use `===` and `!==` instead of `==`  and `!=`. 

https://www.w3schools.com/js/js_best_practices.asp

src/main/webapp/pages/menu.js
```diff
@@ -43,12 +43,15 @@ class Menu {
...
+      var tab = this.createMenuTab(i, isFocused);
       document.querySelector('.mdc-tab-scroller__scroll-content').appendChild(tab);
     }
     const tabBar = new mdc.tabBar.MDCTabBar(document.querySelector('.mdc-tab-bar'));
-    const tabs = document.querySelectorAll('.mdc-tab');
-    tabs[pageName.index].focus();
+    if (pageName != undefined) {
```
cgollner: Same about `!==`

src/main/webapp/pages/usersPortfolio_initPage.js
```diff
@@ -0,0 +1,67 @@
...
+  if (user.name != undefined) {
+    document.getElementById("portfolioName").innerText = user.name;
+  } else {
+    document.getElementById("portfolioName").innerText =
+        user.email.substring(0, user.email.indexOf("@"));
+  }
+  if (user.selfIntroduction != undefined) {
+    document.getElementById("portfolioSelfIntroduction").innerText =
+        user.selfIntroduction;
+  }
```
cgollner, line 57: Should we hide in the `else`?

src/main/webapp/pages/usersPortfolio_initPage.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function setMapWidth() {
+  const availableWidth = window.innerWidth - 370;
+  document.getElementById('mapDisplayer').style.width =
+      availableWidth.toString() + 'px';
+}
+
+function setListHeight() {
+  const availableHeight = window.innerHeight - 520;
```
cgollner: What do these magic values means? 370 and 520. Maybe extract them in a local variable with a descriptive name.

src/main/webapp/pages/usersPortfolio_initPage.js
```diff
@@ -0,0 +1,67 @@
...
+  document.getElementById('listPlaceGuideDisplayer').style.height =
+      availableHeight.toString() + 'px';
```
cgollner: Is `toString()` needed?

src/main/webapp/placeGuides/ListPlaceGuideDisplayer.js
```diff
@@ -42,10 +45,14 @@ class ListPlaceGuideDisplayer {
     return this._listPlaceGuideDisplayerDiv;
   }
 
+  static getPlaceGuideOnListDivId(placeGuideId) {
+    return "placeGuideOnList-" + "{" + placeGuideId + "}";
```
cgollner: Maybe use string templates here as well

src/main/webapp/placeGuides/PlaceGuideOnList.js
```diff
@@ -36,6 +37,44 @@ class PlaceGuideOnList {
...
   }
 
+  static expand(placeGuideId) {
+    const divId = 'placeGuideOnList-' + '{' + placeGuideId + '}';
+    const placeGuideDiv = document.getElementById(divId);
+    placeGuideDiv.querySelectorAll('.folded-placeGuide')[0]
+        .style.display = 'none';
+    placeGuideDiv.querySelectorAll('.card-placeGuide')[0]
+        .style.display = 'block';
+    placeGuideDiv.style.padding = '0px';
```
cgollner, line 47: I'm a bit confused. If it is supposed to expand it, why is it hiding it? Or is it just me not getting CSS? 😅

src/main/webapp/placeGuides/PlaceGuideOnList.js
```diff
@@ -36,6 +37,44 @@ class PlaceGuideOnList {
...
+  }
+
+  static getBlobSrc(blobKey) {
+    const src = new URL('/serve-blob', document.URL);
+    src.searchParams.append('blob-key', blobKey);
+    return src;
+  }
+
+  static generateQueryString(placeGuideProperties) {
+    const esc = encodeURIComponent;
```
cgollner: What is "esc"?

src/main/webapp/placeGuides/PlaceGuideOnList.js
```diff
@@ -36,6 +37,44 @@ class PlaceGuideOnList {
...
+    const query = Object.keys(placeGuideProperties)
+        .map(function(k) {
```
cgollner: Maybe rename "k" to "propertyKey"

src/main/webapp/placeGuides/PlaceGuideOnList.js
```diff
@@ -36,6 +37,44 @@ class PlaceGuideOnList {
...
+          return esc(k) + '=' + esc(placeGuideProperties[k]);
```
cgollner: Wondering if there is a nice URL api from Javascript to add parameters more cleanly?

src/main/webapp/user/User.js
```diff
@@ -30,4 +31,45 @@ class User {
...
+
+  static getUserFromQueryString() {
+    if (window.location.search != '') {
+      const GET = {};
+      const queryString = decodeURI(window.location.search.replace(/^\?/, ''));
+      queryString.split(/\&/).forEach(function(keyValuePair) {
+        const paramName = keyValuePair.replace(/=.*$/, '');
+        const paramValue = keyValuePair.replace(/^[^=]*\=/, '');
+        GET[paramName] = paramValue;
+      });
```
cgollner: This feels like it could be extracted to some "UrlQueryUtils" class so it could be re-used.

src/main/webapp/user/User.js
```diff
@@ -30,4 +31,45 @@ class User {
...
+      const id = GET['id'];
+      let name = GET['name'];
+      if (name == "undefined") {
```
cgollner: Here we are checking `name == "undefined"` but below we check `===`. Let's always use `===`. Or is it different for the name? 
Also is the value really `"undefined"` and not `undefined`? 😮

src/main/webapp/user/UserRepresentation.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  get div() {
+    return this._userRepresentationDiv;
+  }
+
+  static createUserDiv(user) {
+    const userDiv = document.createElement("div");
+    userDiv.style.height = "5px";
+    userDiv.style.width = "50px";
```
cgollner: Shouldn't this be part of some "userDiv" CSS class?

src/main/webapp/user/UserRepresentation.js
```diff
@@ -0,0 +1,72 @@
...
+    const userDivClass = "userRepresentation-" + "{" + user.id + "}";
+    userDiv.setAttribute("class", userDivClass);
+    if (user.imgKey != undefined) {
+      if (user.name != undefined) {
```
cgollner: Again about using `!==`

src/main/webapp/user/UserRepresentation.js
```diff
@@ -0,0 +1,72 @@
...
+        userDiv.appendChild(
+            UserRepresentation.createUserImg(user.imgKey, user.name));
+      } else {
+        userDiv.appendChild(
+            UserRepresentation.createUserImg(user.imgKey, user.email.substring(
+                0, user.email.indexOf("@"))));
```
cgollner: Maybe extract this `user.email.substring(...)` into a helper method with a descriptive name, since we are duplicating it below as well.

cgollner (changes requested):

pom.xml
bori00: I'll remove it. Sorry, it doesn't have to do anything with the project, but since I have started working on the remote desktop I cannot use the default 8080 port anymore, so each time I run a dev server I need to have the port set to something else.

src/main/webapp/pages/usersPortfolio_initPage.js
bori00: I couldn't make the height of the info box dependent on the content unfortunately with the current setup, so it doesn't make any difference to hide it.

src/main/webapp/placeGuides/PlaceGuideOnList.js
bori00: The folded-placeguide and the card-placeguide are different elements, and only one of them is visible all the time. If the card is closed, then folded-placeguide is shown, if the card is expanded, then card-placeguide. I think this explains what happens here. 
Remark that this is not my code, I might have changed the order of the functions only and it appears as if I wrote it.

src/main/java/com/google/sps/data/PlaceGuideQueryType.java
bori00: Done.

src/main/webapp/pages/menu.js
bori00: Changed.

src/main/webapp/pages/menu.js
bori00: Changed.

src/main/webapp/user/User.js
bori00: Yes, all values are strings, because they were sent as strings through the url.

bori00: Changed to ===

src/main/webapp/user/UserRepresentation.js
bori00: Changed.

src/main/webapp/pages/usersPortfolio_initPage.js
bori00: Done, here and everywhere(the constants were used previously on other pages, with the ListRepresentation changes).

src/main/webapp/pages/usersPortfolio_initPage.js
bori00: No, I removed it. Again, I removed it on the other pages as well.

src/main/webapp/placeGuides/ListPlaceGuideDisplayer.js
bori00: Done.

src/main/webapp/user/UserRepresentation.js
bori00: Done.

src/main/webapp/user/UserRepresentation.js
bori00: Sorry, this part remained there from a previous implementation, it is not needed anymore, I removed it completelt.

src/main/webapp/placeGuides/PlaceGuideOnList.js
bori00: Done.

src/main/webapp/placeGuides/PlaceGuideOnList.js
bori00: Not my code, but I assume Dennis used it to make the code shorter. I removed it now.

src/main/webapp/user/User.js
bori00: Done.

src/main/webapp/placeGuides/PlaceGuideOnList.js
bori00: There is, "url.searchParams.append(name, value)", but that would not align with the string parsing method written by Dennis. I would rather leave it as he did it before in this case.

src/main/java/com/google/sps/data/PlaceGuideQueryType.java
```diff
@@ -16,23 +16,33 @@
...
-  CREATED_PRIVATE(false),
-  BOOKMARKED(false),
-  ALL_PUBLIC_IN_MAP_AREA(true),
-  CREATED_ALL_IN_MAP_AREA(true),
-  CREATED_PUBLIC_IN_MAP_AREA(true),
-  CREATED_PRIVATE_IN_MAP_AREA(true);
+  ALL_PUBLIC(/* reuiresCorodinates= */ false, /* requiresUserIdFromRequest= */ false),
+  CREATED_ALL(/* reuiresCorodinates= */ false, /* requiresUserIdFromRequest= */ false),
+  CREATED_PUBLIC(/* reuiresCorodinates= */ false, /* requiresUserIdFromRequest= */ false),
+  CREATED_PRIVATE(/* reuiresCorodinates= */ false, /* requiresUserIdFromRequest= */ false),
```
cgollner: Typo: s/reuiresCorodinates/requiresCoordinates

cgollner (dismissed):

cgollner (approved):

PR #51, opened 2020-09-17
Bug Fix: Editing a guide's location

bori00:
#### Issues
1. When the user attempts to edit an existing guide's location, it doesn't seem to change. 
2. If the user edits a guide whose location is specified through a placeId, after the resubmission the placeId will disappear and the guide will be linked to the previous coordinates, but not to the place

#### Solutions
1. As it turns out, the bug was caused by the fact that instead of the value property of the location input fields of the form, only the value attribute was changed. The attribute change was not reflected in the form's submitted values, and thus the location was not updated. To solve this, I had to change the `.setAttribute('value', value)` statements to `.value = value`. See [this](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute) and [this](https://stackoverflow.com/questions/29929797/setattribute-doesnt-work-the-way-i-expect-it-to) links for a more detailed explanation of the issue. 
2. This one was simply caused by a wrong parameter being passed to the setFormInputValueOrEmpty() function. It accepts a DOM element, but got a DOM element's value instead. I removed the .value getter. 

#### Other changes
While testing this part I realized that the placeName input field is not needed anymore, because it is not displayed to the user anymore and it is not required on the server side either. So I removed it completely.

### DIFF
--- a/src/main/webapp/forms/placeGuideForm.js
+++ b/src/main/webapp/forms/placeGuideForm.js
@@ -56,21 +56,19 @@ function enableSubmission() {
 function updateLocation(position, placeId, placeName) {
   if (placeName != null) {
     document.getElementById(
-        'placeName').setAttribute('value', placeName);
+        'placeName').value = placeName;
     document.getElementById(
-        'placeId').setAttribute('value', placeId);
+        'placeId').value = placeId;
   } else {
     document.getElementById(
-        'placeName').setAttribute('value', '-');
+        'placeName').value = '-';
     document.getElementById(
-        'placeId').setAttribute('value', '');
+        'placeId').value = '';
   }
   document.getElementById(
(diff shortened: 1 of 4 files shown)


src/main/webapp/forms/placeGuideForm.js
```diff
@@ -53,22 +53,18 @@ function enableSubmission() {
  * This function writes in the hidden form inputs the
  * data of the newly chosen location for the placeguide.
  */
-function updateLocation(position, placeId, placeName) {
-  if (placeName != null) {
+function updateLocation(position, placeId) {
+  if (placeId != null) {
```
cgollner: Nit: use `!==`

cgollner (dismissed):

cgollner (approved):

PR #52, opened 2020-09-17
Bug Fix: Geocoding Failures

bori00:
#### Issue
In some cases, PlacesService, which was responsible for geocoding(i.e finding the place with all the details corresponding to an ID) failed, and then the place could not be identified.

image

#### Solution 
I checked the status of PlacesService, and it was OVER_QUERY_LIMIT. (According to some unofficial documentation, one cannot send more than 10 requests to the Places API for geocoding.) To solve this issue, I reorganized the data based on the remark that I don't need the data from the PlacesService directly after fetching the guides from the server: 

- I need it only when the infowindow of the marker is opened
- or when the infobox in the list is expanded

So instead of doing geocoding directly after fetching, for all guides at once, I do it only on-demand, when a DOM event occurs(when the guide is highlighted or when the infobox is expanded. Both happen on clicks). 

To optimize this solution even more, if for one placeguide geocoding was done, I cache the data that I need(currently, it's only the name of the place) in the Location object, so that on the next event I won't repeat the server request. 

However, to make this solution possible, I had to remove the place name field from the folded placeguide card. Instead, I added it to the expanded placeguide card.

### DIFF
--- a/src/main/webapp/placeGuides/Location.js
+++ b/src/main/webapp/placeGuides/Location.js
@@ -26,7 +26,7 @@ class Location {
         });
       }
       const thisLocation = this;
-      this.definePlaceName()
+      return this.definePlaceName()
           .then(newPlaceName => {
             thisLocation._placeName = newPlaceName;
             return newPlaceName;
@@ -36,7 +36,7 @@ class Location {
 
   definePlaceName() {
     const request = {
-      placeId: placeId,
+      placeId: this._placeId,
       fields: ['name'],
     };
     return new Promise(function(resolve, reject) {
--- a/src/main/webapp/placeGuides/Location.js
+++ b/src/main/webapp/placeGuides/Location.js
@@ -3,22 +3,29 @@
  * coordinates or by a place_id, which will be decoded with PlacesService.
  */
 class Location {
-  constructor(position, mapsPlace) {
-    this._mapsPlace = mapsPlace;
-    if (this._mapsPlace != null) {
-      this._position = this._mapsPlace.geometry.location;
-    } else {
-      this._position = position;
-    }
+  constructor(position, placeId) {
+    this._position = position;
+    this._placeId = placeId;
+    this._placeName = undefined;
+    // if (this._mapsPlace != null) {
+    //   this._position = this._mapsPlace.geometry.location;
+    // } else {
+    //   this._position = position;
+    // }
   }
 
   get position() {
     return this._position;
   }
 
-  get mapsPlace() {
-    return this._mapsPlace;
-  }
+  // get placeName() {
+  //   if (this._placeName !== undefined) {
+  //     return this._placeName;
+  //   } else {
+  //
+  //   }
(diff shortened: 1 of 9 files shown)


src/main/webapp/placeGuides/PlaceGuideOnList.js
```diff
@@ -136,27 +122,47 @@ class PlaceGuideOnList {
...
     buttonsContainer.appendChild(expandButton);
     return buttonsContainer;
   }
 
+  static expand(placeGuideId, location) {
+    const divId = 'placeGuideOnList-' + '{' + placeGuideId + '}';
+    const placeGuideDiv = document.getElementById(divId);
+    const placeNameDiv = document.getElementById(
+        PlaceGuideOnList.getPlaceGuidePlaceNameElementId(placeGuideId));
+    if (placeNameDiv !== null && placeNameDiv.innerText === "") {
```
cgollner, line 168: Is this null check needed?

src/main/webapp/placeGuides/Location.js
```diff
@@ -3,40 +3,54 @@
...
+          resolve(undefined);
+        });
+      }
+      const thisLocation = this;
+      return this.definePlaceName()
+          .then((newPlaceName) => {
+            thisLocation._placeName = newPlaceName;
+            return newPlaceName;
+          });
+    }
```
cgollner: If feels a little strange that a getter returns a Promise. In my mind getters usually return just what they have locally. Or is this common in Javascript?

src/main/webapp/placeGuides/PlaceGuideRepository.js
```diff
@@ -18,73 +18,43 @@ class PlaceGuideRepository {
...
-    return this._placeGuides;
-  }
+  // get placeGuides() {
+  //   for (const placeGuideId in this._placeGuides) {
+  //     if (this._placeGuides.hasOwnProperty(placeGuideId)) {
+  //       console.log(this._placeGuides[placeGuideId]);
+  //     }
+  //   }
+  //   return this._placeGuides;
+  // }
```
cgollner: Remove commented code

cgollner (changes requested):

src/main/webapp/placeGuides/PlaceGuideOnList.js
bori00: Yes, because the div for the placename is created only if the placeguide has a placeId(i.e. it is linked to a place to to a point defined by coordinates).

src/main/webapp/placeGuides/PlaceGuideRepository.js
bori00: Done.

src/main/webapp/placeGuides/Location.js
bori00: I converted it to a general function(not a getter).

cgollner (approved):

PR #55, opened 2020-09-23
Bug Fix: unauthorized users editing a guide

bori00:
#### Issue
Although the "edit guide" button was accessible only for the creator of a given guide, the data of the guide was then transmitted though the url(when switching to the createPlaceGuide page with the scope of editing an existing guide), and then any user could edit that guide if they had the link to it, even if they were not authorised(i.e. they are not the creators of the guide). 

#### Solution
1. Client-side changes
Instead of passing all of the data through the url, I modified the code so that only the ID of the guide is transmitted. On the createPlaceGuide page, if an ID is present in the url, then the data of the corresponding guide is fetched from the server and the form is filled with that data. 

2. Server-side changes
To enable getting the data of one particular guide, I added support for a new query type in PlaceGuideServlet: PLACE_GUIDE_WITH_ID.  I also added some tests for the new query. 

Additionally, I now let the user continue editing a guide's data after submission, simply by redirecting to the corresponding url (createPlaceGuide page with the ID parameter of the previously edited guide). This improves the user experience, because the user can see that the changes are indeed saved, and they can also continue editing if they are not satisfied with the result.

### DIFF
--- a/src/main/java/com/google/sps/servlets/PlaceGuideServlet.java
+++ b/src/main/java/com/google/sps/servlets/PlaceGuideServlet.java
@@ -54,6 +54,7 @@ public class PlaceGuideServlet extends HttpServlet {
   public static final String PLACE_GUIDE_QUERY_TYPE_PARAMETER = "placeGuideType";
   public static final String REGION_CORNERS_PARAMETER = "regionCorners";
   public static final String CREATOR_ID_PARAMETER = "creatorId";
+  public static final String PLACE_GUIDE_ID_PARAMETER = "placeGuideId";
 
   private final PlaceGuideRepository placeGuideRepository =
       PlaceGuideRepositoryFactory.getPlaceGuideRepository(RepositoryType.DATASTORE);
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/sps/servlets/PlaceGuideServlet.java
```diff
@@ -72,27 +74,42 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    if (placeGuideQueryType == PlaceGuideQueryType.PLACE_GUIDE_WITH_ID) {
+      long placeGuideId = Long.parseLong(request.getParameter(PLACE_GUIDE_ID_PARAMETER));
+      PlaceGuide placeGuide = placeGuideRepository.getPlaceGuide(placeGuideId);
+      if (placeGuide == null) {
+        throw new IllegalArgumentException("There is no placeguide ith this id: " + placeGuideId);
+      }
+      String currentUserId = UserServiceFactory.getUserService().getCurrentUser().getUserId();
+      PlaceGuideInfo placeGuideInfo = new PlaceGuideInfo(placeGuide, currentUserId);
+      response.setContentType("application/json;");
+      response.getWriter().println(convertToJsonUsingGson(placeGuideInfo));
```
cgollner: I would probably extract the code inside of the conditions to their own methods, to prevent this main `doGet` from growing too much.

cgollner (approved):

src/main/java/com/google/sps/servlets/PlaceGuideServlet.java
bori00: Done.

